PR #1, opened 2020-07-13
service: create draft

ppanero:
**Continue in a new PR #4 to make work easier in diff branches**

*Disclaimer*: The objective of this PR is to merely separate DBs. There are many things "re-used" from records that might need change.

Requires the update of the singleton view in flask-resources

How the DB looks like (Some changes are needed to the db structure indeed):

Screenshot 2020-07-14 at 11 05 04

Created draft:

Screenshot 2020-07-14 at 11 06 24

### DIFF
--- a/invenio_drafts_resources/__init__.py
+++ b/invenio_drafts_resources/__init__.py
@@ -1,0 +1,15 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2020 CERN.
+# Copyright (C) 2020 Northwestern University.
+#
+# Invenio-Drafts-Resources is free software; you can redistribute it and/or
+# modify it under the terms of the MIT License; see LICENSE file for more
+# details.
+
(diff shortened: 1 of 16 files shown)


ppanero:
- Next steps:
    - Redefine model class as a generic draft class. E.g. Since vocabularies, communities, etc. are set as "records", they might also have drafts.
        - Make Service into a parent class, meaning that it will be able to create, etc. drafts for any given datamodel (inject from config)
        - Implemente "Record" drafts in the tests
    - Create a `drafts` module to separate data access layer (`api.py` and others) from business logic (`service/`)
    - Redefine database fields: UUID (same as record it forks from), fork_id (version_id of the forked record), draft_version_id, expire timestamp (optional)
    - *When creating a draft, first it creates the associated record*. This will unlock the possibility to change the PKs etc of the drafts table.

PR #3, opened 2020-07-15
Add Drafts services stubs

### LINKED ISSUE #2: Implement service stubs for Drafts
drafts version of https://github.com/inveniosoftware/invenio-records-resources/issues/23

fenekku:
depends on inveniosoftware/invenio-records-resources#32 (fails until it is merged)

closes #2

Uses some of the code from @ppanero's create branch.


invenio_drafts_resources/resource_unit.py
```diff
@@ -0,0 +1,20 @@
...
+# Copyright (C) 2020 Northwestern University.
+#
+# Invenio-Drafts-Resources is free software; you can redistribute it and/or
+# modify it under the terms of the MIT License; see LICENSE file for more
+# details.
+
+"""Resource Unit."""
+
+
+class IdentifiedDraft:
```
ppanero, line 13: minor: this will probably need some changes to reflect who is it drafting. Probably treated in a following PR (I can do since Im touching database table etc.)

invenio_drafts_resources/resources/draft.py
```diff
@@ -13,6 +13,8 @@
 from flask_resources.context import resource_requestctx
 from flask_resources.resources import ResourceConfig
 
+from ..services import DraftService, DraftVersionService
+
 # TODO: Get rid of them when implementation is done
 STUB_ITEM_RESULT = ({"TODO": "IMPLEMENT ME"}, 200)
```
ppanero, line 19: moderate: remove unused content. I see these got replaced by service calls + todo?

invenio_drafts_resources/resources/file.py
```diff
@@ -14,6 +14,8 @@
 # TODO: expose correctly in flask-resources
 from flask_resources.resources import ResourceConfig
 
+from ..services import DraftFileMetadataService, DraftFileService
+
 # TODO: Get rid of them when implementation is done
 STUB_ITEM_RESULT = ({"TODO": "IMPLEMENT ME"}, 200)
```
ppanero, line 20: same as above.

invenio_drafts_resources/services/draft_file_metadata_service.py
```diff
@@ -0,0 +1,65 @@
+# -*- coding: utf-8 -*-
```
ppanero, line 1: moderate: same file naming issue than in records-resources

invenio_drafts_resources/services/draft_service.py
```diff
@@ -0,0 +1,78 @@
...
+    # Service configuration
+    permission_policy_cls = DraftPermissionPolicy
+    resource_unit_cls = IdentifiedDraft
+
+    # RecordService configuration
+    # TODO: FILL ME!
+
+    # DraftService configuration
+    # TODO: FILL ME!
+    draft_of_resource_unit_cls = IdentifiedRecord
```
ppanero, line 33: I think this might need to be a service class/instance though. e.g. When creating a draft of a non-existing record (new) it will need to create the record + draft (and associate them). Therefore, it would be great to use the service so data validation etc. happens already in the background. For now can be shelved until we have this more clear (just putting my thoughts here)

requirements-devel.txt
```diff
@@ -13,3 +13,4 @@
 #     -e git+git://github.com/mitsuhiko/jinja2.git#egg=Jinja2
 
 -e git+https://github.com/inveniosoftware/flask-resources.git#egg=flask-resources
+-e git+https://github.com/inveniosoftware/invenio-resources.git#egg=invenio-resources
```
ppanero, line 16: major: does not exist? isn't it invenio-records-resources now? (though github keeps the old link working)

minor: flask-resrouces is released right? https://pypi.org/project/flask-resources/ probably we still need unreleased content. Worth adding already to setup.py

ppanero:
I am continuing my work on top of this branch (separate but from it). Therefore, if the comments are treated (mostly the file naming) feel free to merge.

invenio_drafts_resources/resource_unit.py
fenekku: Sure, I leave that up to you (you mean like keeping track of the `draft_of` object, right?)

invenio_drafts_resources/resources/draft.py
fenekku: Yes :broom: !

invenio_drafts_resources/resources/file.py
fenekku: :heavy_check_mark:

invenio_drafts_resources/services/draft_service.py
fenekku: After sleeping on it, I was even wondering if we could just call the `RecordService::create()` method in `publish`, this way that configuration isn't really needed. When implementing this can be changed to something that is more appropriate, so we can wait until that happens.

requirements-devel.txt
fenekku: Yes, probably best to make it match up to the new name. I've also added flask-resources in setup.py and removed it here. :rocket:

fenekku: ah no we need another release of flask-resources. Let's wait to do that change then.

invenio_drafts_resources/services/draft_file_metadata_service.py
fenekku: Done

PR #4, opened 2020-07-16
Continue works on draft creation....

ppanero:
Requires #3 

Next steps:
- [x] Redefine model class as a generic draft class. E.g. Since vocabularies, communities, etc. are set as "records", they might also have drafts.
  - [x] Make Service into a parent class, meaning that it will be able to create, etc. drafts for any given datamodel (inject from config)
  - [x] Implemente "Record" drafts in the tests
- [x] *When creating a draft, first it creates the associated record*. This will unlock the possibility to change the PKs etc of the drafts table.


invenio_drafts_resources/serializers.py
```diff
@@ -0,0 +1,73 @@
...
+            )
+        )
+
+        # TODO: Shall we includ fork_version_id and record_pid in
+        # the serialization?
+        return draft_dict
+
+    def serialize_object(self, obj, response_ctx=None, *args, **kwargs):
+        """Dump the object into a json string."""
+        if obj:  # e.g. delete op has no return body
```
zzacharo, line 61: Just out of curiosity: Where is this used and we have this `if`? Isn't it better to check the `obj` before passing it to the serializer?

ppanero: All reponses call the serializer. And in some cases (E.g. delete) there is no object. However, that is our implementation someone might want to actually return it when deleting.

invenio_drafts_resources/services/draft.py
```diff
@@ -62,51 +56,52 @@ def search(self, querystring, identity, pagination=None, *args, **kwargs):
...
-        # Create record if new
-        # TODO: The `draft_of` class is set in the Draft API
-        # Creating the record class is business logic, but it is
-        # being set in the data access layer
-
-        # Get UUID, and version_id of forked record
-        # Create draft based on that
+        record = self.config.record_cls.create(data=data)
+        pid = self.minter()(record_uuid=record.id, data=record)  # Mint PID
+        marsh_data = self.data_validator().validate(data)
```
zzacharo, line 69: nit: Syntactic sugar: isn't it nicer to have `minter`, `data_validator` as properties in the `RecordService` class so you can write e.g `self.data_validator.validate(data)`?

zzacharo: Should we validate the `data` when we create a record? I remember that in the deposit workflow in the UI, we were having the case of creating an empty record when someone was uploading just a file. In that case, you have available only the file metadata but I assume that your validator it might require more fields I guess...Should we support partial validation when we create a draft/record? @lnielsen @fenekku

ppanero: Hmm I like the syntactic improvement. It would require refactor in invenio-records-resources tho (I can do it, maybe in a different PR - then adapted here)

As for validation. As it is now: It only validates according to the draft schema (I can re-order, to make the validation earlier). What we can do as "partial validation" is: Validate all existing fields (i.e. nothing - or almost nothing -  is required, but if they are they are validated). WDYT?

invenio_drafts_resources/serializers.py
zzacharo: I would guess that when you define a serializer you expect to deal with an object to be serialized...I see more likely people adding their serializer than overriding the resource methods :) But no strong opinion here!

invenio_drafts_resources/services/draft.py
zzacharo: I agree! I think when creating/saving a draft the partial behaviour would be to validate only what is passed in the incoming data through the marshmallow schema. Then, when you publish/edit you require full validation!

zzacharo: For the syntactic improvement we can create an issue to address later if we like it :)

ppanero: Issues opened: 
- https://github.com/inveniosoftware/invenio-records-resources/issues/36
- https://github.com/inveniosoftware/invenio-drafts-resources/issues/6

> I agree! I think when creating/saving a draft the partial behaviour would be to validate only what is passed in the incoming data through the marshmallow schema. Then, when you publish/edit you require full validation!

Will refactor this after Im done gluing the service and views

invenio_drafts_resources/serializers.py
ppanero: Hmm lets make a vote in some standup and see whats best?

invenio_drafts_resources/services/draft.py
ppanero: @zzacharo I have implemented the glue with the views in #7 I will tackle the data validation order in that PR to avoid this one hanging forever. I have set the `FIXME` there :)

invenio_drafts_resources/resources/draft.py
```diff
@@ -50,7 +50,9 @@ def create(self, *args, **kwargs):
         """Create an item."""
         data = resource_requestctx.request_content
         identity = g.identity
-        return self.service.create(data, identity), 200
+        id_ = resource_requestctx.route["pid_value"]
```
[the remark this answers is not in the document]
zzacharo: Hmmm, sorry I missed that...what is the idea behind this? I mean this is used for `create` new and `edit`?

tests/conftest.py
```diff
@@ -50,15 +89,31 @@ def create_app(instance_path):
...
+        record_bp = RecordResource().as_blueprint("record_resource")
+        draft_bp = DraftResource().as_blueprint("draft_resource")
+
+        app.register_blueprint(record_bp)
+        app.register_blueprint(draft_bp)
         yield app
 
 
+@pytest.fixture(scope="module")
+def draft_service(instance_path):
```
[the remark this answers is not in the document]
ppanero: Oops, copy-paste error thanks!

invenio_drafts_resources/resources/draft.py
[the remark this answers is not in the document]
ppanero: Yes, it is. The config `list_route` is `list_route = "/records/<pid_value>/draft"`

invenio_drafts_resources/config.py
```diff
@@ -8,3 +8,9 @@
 # details.
 
 """Invenio-Drafts-Resources Configuration."""
+
+DRAFTS_RESOURCES_LINK_URLS = {
+    'record': '{base}/records/{pid}',
+    'records': '{base}/records/',
+    'draft': '{base}/records/{pid}/draft/',
+}
```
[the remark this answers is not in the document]
ppanero: Shelvable? I would not tie UI and API even if its for config. So something different should be used, centralized but not tight to any of them (nor UI nor API). Therefore I think its a bigger task and gets a bit out of the scope of this PR/task.

I opened https://github.com/inveniosoftware/invenio-drafts-resources/issues/8 for it.

invenio_drafts_resources/serializers.py
ppanero: Shelving to allow continuation of the work in #7 issue -> https://github.com/inveniosoftware/invenio-drafts-resources/issues/9

invenio_drafts_resources/services/draft.py
```diff
@@ -30,83 +27,55 @@ class DraftServiceConfig(RecordServiceConfig):
...
+        """Create a draft and the associated record (new)."""
         self.require_permission(identity, "create")
-        # Create record if new
-        # TODO: The `draft_of` class is set in the Draft API
-        # Creating the record class is business logic, but it is
-        # being set in the data access layer
-
-        # Get UUID, and version_id of forked record
-        # Create draft based on that
+        record = self.config.record_cls.create(data=data)
```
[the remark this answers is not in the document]
ppanero: Discussed IRL: Not eager.

invenio_drafts_resources/services/draft.py
```diff
@@ -30,83 +27,55 @@ class DraftServiceConfig(RecordServiceConfig):
...
+    def _index_draft(self, draft):
+        indexer = self.indexer()
+        if indexer:
+            indexer.index(draft)
 
-    def create(self, data, identity):
-        """Create a draft."""
-        # TODO: IMPLEMENT ME!
-        # # Check permissions: Can create draft of said record/resource
+    def create_new(self, data, identity):
```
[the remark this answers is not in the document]
ppanero: Discussed IRL:

- Naming of the service --> `RecordDraftService`
- `create_new` --> `create`
- `create_from` --> `edit` (creates a new draft)
- `edit` cannot be called `update` because the draft can still be *updated*

invenio_drafts_resources/resources/record.py
```diff
@@ -0,0 +1,31 @@
...
+    def __init__(self, service=None, *args, **kwargs):
+        """Constructor."""
+        super(RecordResource, self).__init__(*args, **kwargs)
+        self.service = service or DraftService()
+
+    def create(self, *args, **kwargs):
+        """Create an item."""
+        data = resource_requestctx.request_content
+        identity = g.identity
+        return self.service.create_new(data, identity), 200
```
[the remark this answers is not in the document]
ppanero: Well.... it actually disappeared now 🔥

invenio_drafts_resources/services/draft.py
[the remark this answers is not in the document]
ppanero: Added fixme, treating in #7 because it requires a lot of changes

PR #7, opened 2020-07-17
draft creation: adds the logic needed to use the draft creation on both endpoints

ppanero:
Requires #4

Closes https://github.com/inveniosoftware/invenio-app-rdm/issues/211 🎊 

Adds the logic needed to use the draft creation on both endpoints:
- `/records`
- `/records/<pid_value>/draft`

**Problem**:

The resource unit class is not common between them and the serializer needs to be aware of both. We can simply use `IdenfiedRecord` even if it is for a draft. In the end the content is the same. If needed, for better naming, we can change it to `IdentifiedUnit` (The class name).


tests/test_draft_resource.py
```diff
@@ -0,0 +1,97 @@
...
+    for field in fields_to_check:
+        assert field in response_fields
+
+    recid = response.json['pid']
+
+    # Test the associated record is created
+    response = client.get(
+        "/records/{}".format(recid),
+        headers=HEADERS,
+        data=json.dumps(input_draft)
```
fenekku, line 41: normal: don't think this is needed

tests/test_draft_resource.py
```diff
@@ -0,0 +1,97 @@
...
+    response_fields = response.json.keys()
+    fields_to_check = ['pid', 'metadata', 'revision',
+                       'created', 'updated', 'links']
+
+    for field in fields_to_check:
+        assert field in response_fields
+
+    recid = response.json['pid']
+
+    # Test the associated record is created
```
fenekku, line 37:
```suggestion
    # Test the draft is created
```

tests/test_draft_resource.py
```diff
@@ -0,0 +1,97 @@
...
+    orig_title = input_record['title']
+    input_record['title'] = "Edited title"
+
+    response = client.post(
+        "/records/{}/draft".format(recid),
+        headers=HEADERS,
+        data=json.dumps(input_record)
+    )
+
+    assert response.status_code == 200
```
fenekku, line 75: moderate: I think we want 201 (Created)

tests/test_draft_resource.py
```diff
@@ -0,0 +1,97 @@
...
+    for field in fields_to_check:
+        assert field in response_fields
+
+    assert response.json['metadata']['title'] == input_record['title']
+
+    # Check the actual record was not modified
+    response = client.get(
+        "/records/{}".format(recid),
+        headers=HEADERS,
+        data=json.dumps(input_record)
```
fenekku, line 89: remove `data=...`

tests/test_draft_resource.py
ppanero: 🔥

tests/test_draft_resource.py
ppanero: why exactly? if removed since there are many arguments possible to the function it will fail due to positioning (it's not just `data, headers` either). TBH, for clarity I think it helps to have it and it does not affect in any case...

tests/test_draft_resource.py
ppanero: Yes thanks!

ppanero:
Tests failing due to 201 needed in records-resources (@zzacharo is on it). As for the `IdentifiedDeposit`, I would shelve and rename once agreed. I am not convinced of using Drafts and Deposits at the same time (it is a change in meaning from old deposit and is confusing, even for me, so I guess even more for users or devs not present in this part of development)

invenio_drafts_resources/drafts/models.py
```diff
@@ -26,13 +27,20 @@ class DraftMetadataBase(Timestamp):
...
+
+    fork_id = db.Column(UUIDType)
     """Draft identifier, it is the same than the record it is draft of"""
 
-    fork_version_id = db.Column(db.Integer, primary_key=True)
+    fork_version_id = db.Column(db.Integer)
     """Version id of the record it is draft of."""
 
-    version_id = db.Column(db.Integer, primary_key=True)
+    version_id = db.Column(db.Integer, nullable=False)
```
zzacharo, line 43: Won't we need to have as a PK id + version_id?

invenio_drafts_resources/resource_units.py
```diff
@@ -3,19 +3,24 @@
...
+    """Resource unit representing pid + Record data clump.
+
+    Since a draft can be created for a non-existing record,
+    it might not have a PID.
+    """
+
+    def __init__(self, pid=None, record=None):
+        """Initialize the record state."""
+        self.id = pid.pid_value if pid else None
         self.pids = [pid]
```
zzacharo, line 25: If pid is None you will end up with `[None]`. Is this intentional?

invenio_drafts_resources/services/draft.py
```diff
@@ -51,30 +51,28 @@ def _index_draft(self, draft):
 
     def create(self, data, identity):
         """Create a draft and the associated record (new)."""
```
zzacharo, line 53: Maybe we need to update the docstring as we don't create a new record on the draft creation :)

invenio_drafts_resources/drafts/models.py
ppanero: Nope, because we are not creating the record eagerly anymore. Therefore we cannot always use the UUID (id) of the record in the draft. As a consequence the draft has its own UUID, which is PK. Then the version_id becomes just that, a version_id. The same reason applies to the proceeding question.

invenio_drafts_resources/resource_units.py
ppanero: Because of the above. Since we do not want to mint drafts, they do not have a PID if they are for a non-existing record. If the record exists, they do have a pid. Therefore `[None]` is okay when dealing with drafts.

invenio_drafts_resources/services/draft.py
ppanero: True, done :)

invenio_drafts_resources/resource_units.py
zzacharo: I was just thinking that if you have `pids=[None]` then any code that iterates through it, it must check `if pid is not None`. However, we could just check before if pid when we assign the value to attribute `self.pids`. An empty array would be more developer friendly :) WDYT?

ppanero: You are totally correct. Adding that in a separate PR :)

invenio_drafts_resources/drafts/models.py
zzacharo: I am just thinking that, uuid even unlikely it could end up to give back an existing value:) Therefore, I think that was the case even before, we were using the tuple (uuid, version_id) so we can be sure that PK is unique when we create a draft/record :) Just for clarification, I don't mean to use any record related uuid or version ids!

ppanero: **I think there are two questions here:**

*Probability of collision of UUID*

Firstly if this is a concern, we need to change the`records` model, since it only uses a PK of UUID (no version): https://github.com/inveniosoftware/invenio-records/blob/master/invenio_records/models.py#L44

Now, statistically speaking. At first glance with 32 chars I would say the amount of produced strings is 36^32 (36 available chars power 32 chars chosen). Now google says that is 6.3340287e+49, a number I'm not even able to pronounce in terms of billions, millions, etc. I *hardly* think we will get there, and if so it means our repository is bigger than google's web index (supposed to have 130 trillion IIRC, which is around power 18).
If every person on the planet generates a new UUID4 every second, we’d expect a collision to occur after about 10 years.

*Using `version_id` for a PK*:

The problem with this statement is that in *any* case the uuid and the version id are independent. Even if the uuid collides, which I think is more than proven it won't, it can still collide with version_id. i.e two new drafts for non-existing records will get version id 1, if in addition the uuid collide... well we have a PK collision of uuid + version id, since both attributes are the same.

zzacharo: Thanks for the detailed explanation :) Very interesting statistics 😅  And yes, you are right about the version id ;)

tests/test_draft_resource.py
fenekku: This was just the same issue as above : we are passing a payload to a call to the GET endpoint. That endpoint doesn't expect or need this payload - I think it was just a copy paste leftover...  :)

ppanero: True, I misunderstood you. I thought you meant in general... Tackled in #10

PR #16, opened 2020-07-28
tests: remove unnecesary fixtures

ppanero:


tests/test_draft_resource.py
```diff
@@ -20,8 +20,7 @@
 HEADERS = {"content-type": "application/json", "accept": "application/json"}
 
 
-def test_create_draft_of_new_record(client, draft_service, input_draft,
-                                    fake_identity):
```
fenekku, line 24: :fire:

PR #23, opened 2020-08-03
services, draft: fix minter order to add recid field to db content

ppanero:
Previously the minting was happening after the the DB and therefore the `recid` was only present in the indexed content. Later on, when publishing the record was re-indexed without the `recid` (since at resolve time from db it was not there). This was causing problems in the search.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -70,7 +70,7 @@ install_requires = [
     "invenio-records-resources>=0.3.2,<0.4.0",
     "invenio-accounts>=1.3.0",
     "invenio-files-rest>=1.2.0",
-    "invenio-records-permissions>=0.8.0",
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -70,7 +70,6 @@
     "invenio-records-resources>=0.3.2,<0.4.0",
     "invenio-accounts>=1.3.0",
     "invenio-files-rest>=1.2.0",
-    "invenio-records-permissions>=0.8.0",
```
fenekku, line 73: Was this meant to be deleted? It is still used :) (I know it will be installed via invenio-records-resources, but that's perhaps too implicit and indirect yet...)

ppanero: fair enough, I took it out cuz it was coming from records-resources. Will put it back int.

PR #24, opened 2020-08-04
services: split records and drafts indexing

### LINKED ISSUE #21: service: support replace of record when publish draft
When publishing a record there are several things happening:
1- PID status change from `NEW` to `REGISTERED`
2- Draft deletion, Record creation in DB
3- Draft deletion, Record indexing in ES (This last step has a problem).

The first implementation was as follow:

``` python
        if self.indexer:
            self.indexer.index(record)
            self.indexer.delete(draft)
```

This was silently failing: The record was sent to index, however, since the UUID was the same it would simply not. Then the draft (same UUID than the record) would be deleted. End result: No draft nor record indexed in ES. The test is checking on `/records/<pid_value>` if the record was published, but this operation does not access ES but DB. Therefore, *it works :tm:*.

The correct option would be:

``` python
        if self.indexer:
            self.indexer.delete(draft)
            self.indexer.index(record)
```

The problem now is a version conflict:
1- The UUID is the same. The draft has UUID + version 1.
2- The draft is deleted (but not really, just marked as is). For more info see [some more info about deletes section](https://www.elastic.co/blog/elasticsearch-versioning-support).
3- The garbage colector takes 1 minutes to run.
4- The indexing happens before the above deleted draft (same UUID) is collected. Therefore, it is trying to index UUID + version_0 when UUID + version_1 (or version_0, does not matter since the one trying to be indexed it's not > 0). This results in a version conflict.

Solutions:

- Use different indexes (**My preferred option**)
- Implement document replace/update
- Implement some strategy to sync with the garbage collector of ES (**highly not recommended**)

Should be tackled along with versioning.

ppanero:
Closes #21 

Example of how the documents progress (first draft, then record)
```bash
(containers) ~/Workspace/inveniosw/test/containers curl -XGET http://localhost:9200/_cat/indices\?v
health status index                                     uuid                   pri rep docs.count docs.deleted store.size pri.store.size
yellow open   communities-communities-v1.0.0-1596543514 r9B_iylrR3uFWRQlRSHdIg   1   1          0            0       230b           230b
yellow open   drafts-record-v1.0.0-1596543514           HfaxiFN9SE-SG-aSD_ytfQ   1   1          0            0       230b           230b
yellow open   records-record-v1.0.0-1596543514          -ZvrDKz3QCKSPvdkyKP-hQ   1   1          0            0       230b           230b
(containers) ~/Workspace/inveniosw/test/containers curl -XGET http://localhost:9200/_cat/indices\?v
health status index                                     uuid                   pri rep docs.count docs.deleted store.size pri.store.size
yellow open   communities-communities-v1.0.0-1596543514 r9B_iylrR3uFWRQlRSHdIg   1   1          0            0       230b           230b
yellow open   drafts-record-v1.0.0-1596543514           HfaxiFN9SE-SG-aSD_ytfQ   1   1          1            0      9.7kb          9.7kb
yellow open   records-record-v1.0.0-1596543514          -ZvrDKz3QCKSPvdkyKP-hQ   1   1          0            0       230b           230b
(containers) ~/Workspace/inveniosw/test/containers curl -XGET http://localhost:9200/_cat/indices\?v
health status index                                     uuid                   pri rep docs.count docs.deleted store.size pri.store.size
yellow open   communities-communities-v1.0.0-1596543514 r9B_iylrR3uFWRQlRSHdIg   1   1          0            0       230b           230b
yellow open   drafts-record-v1.0.0-1596543514           HfaxiFN9SE-SG-aSD_ytfQ   1   1          0            0        3kb            3kb
yellow open   records-record-v1.0.0-1596543514          -ZvrDKz3QCKSPvdkyKP-hQ   1   1          1            0      9.7kb          9.7kb
```


.travis.yml
```diff
@@ -34,8 +34,8 @@ env:
     - ES_HOST=127.0.0.1
     - SQLALCHEMY_DATABASE_URI="postgresql+psycopg2://postgres@localhost:5432/invenio"
   matrix:
-    # - REQUIREMENTS=lowest ES_URL=$ES7_DOWNLOAD_URL
-    # - REQUIREMENTS=release ES_URL=$ES7_DOWNLOAD_URL
+    - REQUIREMENTS=lowest ES_URL=$ES7_DOWNLOAD_URL
+    - REQUIREMENTS=release ES_URL=$ES7_DOWNLOAD_URL
```
fenekku, line 38: These also need the `EXTRAS`

invenio_drafts_resources/services/draft.py
```diff
@@ -56,6 +58,13 @@ def resolver(self):
             registered_only=False
         )
 
+    @property
+    def indexer(self):
+        """Factory for creating an indexer instance."""
+        return self.config.indexer_cls(
+            record_to_index=self.config.record_to_index
+        )
```
fenekku, line 66: We can eventually push this up to invenio-records-resources, but I do prefer starting with it here :+1:

invenio_drafts_resources/services/search.py
```diff
@@ -0,0 +1,28 @@
...
+    """Return the index alias for records and drafts based objects.
+
+    Note: document type is deprecated and therefore returned as None
+    """
+    if isinstance(record, DraftBase):
+        return 'drafts', None
+    if isinstance(record, RecordBase):
+        return 'records', None
+    else:
+        return None, None
```
fenekku, line 28: Given the discussion about `'schemas'` yesterday, this would have to be changed to look for the `'schemas'` field right? The underlying problem being that there can be multiple versions of the draft schema as we evolve in time. 

I am fine with keeping this as is and investigating in another task.

tests/conftest.py
```diff
@@ -96,7 +108,9 @@ def app_config(app_config):
     https://github.com/inveniosoftware/invenio-records-permissions/issues/51
     """
     app_config["RECORDS_REST_ENDPOINTS"] = {}
-
+    app_config["INDEXER_DEFAULT_DOC_TYPE"] = "testrecord"
+    app_config["INDEXER_DEFAULT_INDEX"] = TestSearch.Meta.index
+    app_config["SEARCH_MAPPINGS"] = ['drafts', 'records']
```
fenekku, line 113: minor: not necessary right? (we don't do it in invenio-records-resources)

.travis.yml
ppanero: Nice catch thanks!

invenio_drafts_resources/services/draft.py
ppanero: Shelved: Yes! will be refactored also along a small fix in `invenio-indexer`

tests/conftest.py
ppanero: I did it for completeness, this avoids an issue if some other app instantiated indexes (with same alias might still somehow class).

invenio_drafts_resources/services/search.py
ppanero: Shelving it: got numerous feedback from lars today. Will create an issue (working on it now) with all the points to change.

PR #25, opened 2020-08-06
versioning: record/draft versioning

ppanero:
- [x] Requires #24 
- [x] Requires https://github.com/inveniosoftware/invenio-records-resources/pull/63 (and release of `invenio-records-resources`, can be a patch since the change is backcompat).

- Adds `conceptrecid` to the Marshmallow schema
- Adds `conceptrecid` and `recid` to the ES mappings (mocked for testing)
- Creates a new minter: This will create a `RESERVED` PID for the `conceptrecid`. It will mint a `RESERVED` PID (+UUID) for the draft. It will link them (relate them).
- New class `PIDManager`: https://github.com/inveniosoftware/invenio-drafts-resources/issues/33
- Partial data validation: Thanks to PR #63 in invenio-records-resources data validation can be done partially. Meaning required fields are not `required`. This is what is needed for draft validation *validate only the fields that were given*. Therefore, now the draft service can use a single `data_validator`, even better: the default one from invenio-records-resources.
- Lots of improvements form https://github.com/inveniosoftware/invenio-drafts-resources/issues/28

**Questions**:
- Should we add `conceptrecid` to the `pids` attribute list in the resource unit.

Closes https://github.com/inveniosoftware/invenio-drafts-resources/issues/28
Closes https://github.com/inveniosoftware/invenio-drafts-resources/issues/30
Closes https://github.com/inveniosoftware/invenio-app-rdm/issues/219
Closes https://github.com/inveniosoftware/invenio-drafts-resources/issues/37


invenio_drafts_resources/services/draft.py
```diff
@@ -71,85 +62,141 @@ def draft_cls(self):
...
-        return self.resource_unit(pid=pid, record=record)
+        return self.resource_unit(pid=pid, record=draft, links=None)
+
+    def update_draft(self, id_, data, identity):
+        """Replace a record."""
+        # TODO: etag and versioning
+        pid, draft = self.pid_manager.resolve(id_, draft=True)
+        # Permissions
+        self.require_permission(identity, "update", record=draft)
+        validated_data = self.data_validator.validate(data, partial=True)
```
zzacharo, line 81: 🚀

invenio_drafts_resources/services/draft.py
```diff
@@ -71,85 +62,141 @@ def draft_cls(self):
...
+
+    def new_version(self, id_, identity):
+        """Create a new version of a record."""
+        self.require_permission(identity, "create")
+        # Get record
+        pid, record = self.pid_manager.resolve(id_)
+        # Validate and create a draft, register PID
+        data = record.dumps()
+        # Validate against record schema
+        validated_data = self.data_validator.validate(data, partial=True)
```
zzacharo, line 186: Isn't it here better to have strict validation as we start from a record?

invenio_drafts_resources/services/minter.py
```diff
@@ -0,0 +1,52 @@
...
+    data['conceptrecid'] = conceptrecid.pid_value
+    return conceptrecid
+
+
+# By default both recid and conceptrecid will be RESERVED
+RecordIdProviderV2.default_status_with_obj = PIDStatus.RESERVED
+
+
+def versioned_recid_minter_v2(record_uuid, data):
+    """Rserve the Concept RECID and create the RECID."""
```
zzacharo, line 36: small typo :)

invenio_drafts_resources/services/draft.py
ppanero: I actually wonder if we should validate it? It's a nice-to-do but the record was already validated when publishing... WDYT?

zzacharo: It makes sense not to validate :)

invenio_drafts_resources/services/draft.py
```diff
@@ -71,85 +62,138 @@ def draft_cls(self):
...
         """Create a draft for a new record.
 
         It does not eagerly create the associated record.
         """
         self.require_permission(identity, "create")
-        validated_data = self.draft_data_validator.validate(data)
+        validated_data = self.data_validator.validate(data, partial=True)
         rec_uuid = uuid.uuid4()
-        pid = self.minter(record_uuid=rec_uuid, data=validated_data)
+        pid = self.pid_manager.mint(record_uuid=rec_uuid, data=validated_data)
```
fenekku, line 100: :100:  Much better than the minters that we used to have.

invenio_drafts_resources/services/draft.py
```diff
@@ -71,85 +62,138 @@ def draft_cls(self):
...
-        pid, record = self.resolve(id_)
+        pid, record = self.pid_manager.resolve(id_)
         self.require_permission(identity, "create")
-        validated_data = self.draft_data_validator.validate(data)
-        draft = self.draft_cls.create(validated_data, id_=record.id,
+        validated_data = self.data_validator.validate(data, partial=True)
+        self._patch_data(record, validated_data)
+        draft = self.draft_cls.create(record.dumps(), id_=record.id,
                                       fork_version_id=record.revision_id)
         db.session.commit()  # Persist DB
```
fenekku, line 126: Question: Above in `update_draft` we do `draft.commit()`, shouldn't we do the same here?

invenio_drafts_resources/services/draft.py
```diff
@@ -71,85 +62,138 @@ def draft_cls(self):
...
-        data = draft.dumps()
+        pid, draft = self.pid_manager.resolve(id_, draft=True)
         # Validate against record schema
-        validated_data = self.data_validator.validate(data)
-        # Use same UUID
-        record = self.record_cls.create(validated_data, id_=draft.id)
-        pid.register()
+        validated_data = self.data_validator.validate(draft.dumps())
+        # Publish it
+        if self.pid_manager.is_published(pid):  # Then we publish a revision
```
fenekku, line 164: note to self: why the disctinction?

invenio_drafts_resources/services/draft.py
```diff
@@ -71,85 +62,138 @@ def draft_cls(self):
...
+        rec_uuid = uuid.uuid4()
+        pid = self.pid_manager.mint(
+            record_uuid=rec_uuid,
+            data=data
+        )
+        draft = self.draft_cls.create(
+            data=data,
+            id_=rec_uuid
+        )
+        db.session.commit()  # Persist DB
```
fenekku, line 194: As above: `draft.commit()` ?

invenio_drafts_resources/services/minter.py
```diff
@@ -0,0 +1,52 @@
...
+        pid_value=str(RecordIdProviderV2.generate_id()),
+        status=PIDStatus.RESERVED,
+    )
+
+    data['conceptrecid'] = conceptrecid.pid_value
+    return conceptrecid
+
+
+# By default both recid and conceptrecid will be RESERVED
+RecordIdProviderV2.default_status_with_obj = PIDStatus.RESERVED
```
fenekku, line 32: Shelve: This is a global assignment that may have unintended consequences + makes it hard to test. We should revisit this eventually. Made issue: https://github.com/inveniosoftware/invenio-pidstore/issues/133

invenio_drafts_resources/services/pid_manager.py
```diff
@@ -0,0 +1,100 @@
...
+
+
+class PIDManager:
+    """PID manager."""
+
+    default_config = PIDManagerConfig
+
+    def __init__(self, config=None):
+        """Constructor."""
+        self.config = config or self.default_config
```
fenekku, line 37: This is definitely a pattern now :smile_cat: 

We will have to see how class configurations (versus instances) fare overtime.

tests/conftest.py
```diff
@@ -152,10 +162,14 @@ def app(app):
...
             service=_draft_service()
         ).as_blueprint("draft_action_resource")
+        draft_version_bp = DraftVersionResource(
+            service=_draft_service()
+        ).as_blueprint("draft_version_resource")
 
         app.register_blueprint(record_bp)
         app.register_blueprint(draft_bp)
         app.register_blueprint(draft_action_bp)
+        app.register_blueprint(draft_version_bp)
```
fenekku, line 172: Shelve: We will need to think about a better way to incorporate actions or one-off endpoint in flask-resources... Clearly having to create a new resource each time highlights that we are missing something.

tests/test_draft_resource.py
```diff
@@ -75,59 +116,55 @@ def test_create_draft_of_existing_record(client, record_service,
...
     )
 
     assert response.status_code == 201
-    recid = response.json['pid']
+    _assert_single_item_response(response)
+    assert response.json['revision'] == 0
+    recid_2 = response.json['metadata']['recid']
 
-    # Publish it
+    # Publish it to check the increment in reversion
```
fenekku, line 161:
```suggestion
    # Publish it to check the increment in new version
```

tests/test_draft_resource.py
```diff
@@ -75,59 +116,55 @@ def test_create_draft_of_existing_record(client, record_service,
...
-
-    assert response.status_code == 200
-
-    response_fields = response.json.keys()
-    fields_to_check = ['pid', 'metadata', 'revision',
-                       'created', 'updated', 'links']
-
-    for field in fields_to_check:
-        assert field in response_fields
+    assert response.json['metadata']['recid'] == recid_2 != recid
```
fenekku, line 169: Nice! For some reason it didn't click for me that a triple equality comparison is possible even if I use `a <= b < c` all the time.

invenio_drafts_resources/resources/draft.py
```diff
@@ -73,6 +71,11 @@ class DraftVersionResourceConfig(ResourceConfig):
     """Draft resource config."""
 
     list_route = "/records/<pid_value>/versions"
+    response_handlers = {
+        "application/json": RecordResponse(
+            RecordDraftJSONSerializer(schema=RecordDraftSchemaJSONV1)
```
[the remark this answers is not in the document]
ppanero: Fully correct, fixed :)

invenio_drafts_resources/resources/draft.py
```diff
@@ -119,8 +124,9 @@ def __init__(self, service=None, *args, **kwargs):
...
     def create(self, *args, **kwargs):
         """Any POST business logic."""
+        # FIXME: Implement as CMD patter loaded from config
         if resource_requestctx.route["action"] == "publish":
             identity = g.identity
             id_ = resource_requestctx.route["pid_value"]
-            return self.service.publish(id_, identity), 200
-        return {}, 200
+            return self.service.publish(id_, identity), 202
+        return {}, 202
```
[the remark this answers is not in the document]
ppanero: I would leave as 202. It is true for now it is not an async task. But once the command pattern is there for services, it will be. My point being it will be soon, as soon as we start adding more complexity to it.

invenio_drafts_resources/services/minter.py
ppanero: I could not agree more. IMO this should be an instance attribute, not class. The main issue is that refactoring that would be a huge breaking change in `invenio-pidstore`, which is at the core of every Invenio repository... And IIRC, the idea for now is not to touch those, or keep to a minimum. 

I actually had to do a PR there to add this, but had to keep it back compat...

I think all this we can tackle once we have an RDM 1.0.0 out :) 🚀

invenio_drafts_resources/services/pid_manager.py
ppanero: Yeah, I think we are sort of rendering useless `config.py`... on the other hand, users can still define classes, which I think is clearer than dicts (Biased developer opinion there hehe)

tests/conftest.py
ppanero: We might want to provide yes-another-level of abstraction. Sort of `as_blueprint(actions=True, version=True)` or something like that.

tests/test_draft_resource.py
ppanero: TBH, I didn't click on that two. I just said "Oh, would this work?" I checked online, it did... voila! It's nice and might save quite some lines :)

tests/test_draft_resource.py
ppanero: Nice catch hehe

invenio_drafts_resources/services/draft.py
ppanero: 😃

invenio_drafts_resources/services/draft.py
```diff
@@ -31,32 +28,26 @@ class RecordDraftServiceConfig(RecordServiceConfig):
...
-    draft_data_validator = MarshmallowDataValidator(
-        schema=DraftMetadataSchemaJSONV1
-    )
 
 
 class RecordDraftService(RecordService):
-    """Draft Service interface."""
+    """Draft Service interface.
+
+    This service includes verioning.
```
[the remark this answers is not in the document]
ppanero: Merci! sorry for that 😅

invenio_drafts_resources/resources/draft.py
```diff
@@ -92,8 +95,10 @@ def search(self, *args, **kwargs):
 
     def create(self, *args, **kwargs):
         """Create an item."""
-        # TODO: IMPLEMENT ME!
-        return self.service.create()
+        identity = g.identity
+        id_ = resource_requestctx.route["pid_value"]
+
+        return self.service.new_version(id_, identity), 201
```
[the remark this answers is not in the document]
ppanero: Yes, I agree with this. The response creation of flask (`make_response`) does the same: First arguments that are always provided, if you have extra they go at the end :)

https://github.com/inveniosoftware/invenio-drafts-resources/issues/39

invenio_drafts_resources/services/draft.py
ppanero: This is mostly for the workflow. We have three cases:

- Record does not exists: We create a draft, then later it can be published. In this case *is_published == False*, then we need to register the `recid` (and `conceptrecid`). In addition, we have to create the record itself in the db.

- Record exists, we create a new version: In this case, we create a draft with a different `recid` (same `conceptrecid`). That means that when publishing the new version, the record *is_published == False*. Then we have to register (just) the `recid` and create the record itself in db.

- Record exists, we want to edit it: This means, changing metadata (what happens when `service.edit()`). This just need to update the json blob of the current record, there is no need to register pids or anything else. 

The distinction in the end comes from: `new version` happens when the files of the record change, `new revision` happens when the metadata changes. However, if the user decides to click `edit` won't be able to change files, and if clicks `new version` but doesn't change files it won't be able to publish.

invenio_drafts_resources/services/draft.py
ppanero: Since the delete and create do something like `db.session.delete(record.model)`, you need to `db.session.commit()` the session itself. The operation just marks the SQLAlchemy model for deletion, it's persisted when the session is commited.

The `record.commit` is basically a necessary evil because SQLAlchemy cannot automatically  track changes to JSON (blob) columns. So it's just merging on it. Maybe @slint (who gave me this explanation) or @lnielsen can give more details.

My understanding is:
- Delete and Create sort of "execute plain delete and insert statements" (under the hood obviously). Therefore, they are full actions that need to be just commited. They can mark the full entity (self.model)
- To update we have to nest + merge, so we can "run an specific update statement on a blob".  Mainly because it  cannot track the `self.model.json` (cuz it is a blob (?)).  So we need to merge that specific field instead of update + commit session. 

I think the confusion might be due to naming. In the end the `record.commit` is only called on updates... so it might as well be part of said function?

ppanero: Example given by @slint 
```python
from sqlalchemy import flag_modified

class Record(db.Model):

    id = db.Column(db.UUID, primary_key=True, default=uuid.uuid4)
    json = db.Column(db.JSON)
    title = db.Column(db.Str)


# Just creates an SQLAlchemy model object
record = Record(json={'foo': 'bar'}, title='some title')

# adds the model to the session (you can find it in `db.session.new`)
db.session.add(record)

# Commits the current session state to the DB. SQLAlchemy checks what things
# you have in `db.session.{new,dirty,deleted}` and then generates the appropriate
# SQL statements. e.g. now it will send:
#    INSERT INTO records (json, title) VALUES ({'foo': 'bar'}, 'some title');
db.session.commit()

# Since we set directly the field, there is a "__set__" method called on the "title" member
# which marks the object as "dirty" and adds it to the session. After this you can
# check in `db.session.dirty` and you'll find the `record` object
record.title = 'something else'
# In the same fashion as above, this will generate:
#    UPDATE records WHERE id = '<record-uuid>' SET title = 'something else'
db.session.commit()


# This will not trigger the `json.__set__` since we're modifying the dictionary
# value directly... So if you check in `db.session.dirty` you won't find the `record` model
record.json['foo'] = 'baz'
# If instead we did something like:
# 
#    record.json = {'foo': 'baz'}
# 
# ...it would go through `json.__set__` and thus mark the object as "dirty"

# To achieve what we want we flag the object as dirty manually (this is what `record.commit()` does in Invenio):
flag_modified(record)

# This will send the update as expected:
#     UPDATE records WHERE id = '...' SET json = {'foo': 'baz'}
db.session.commit()
```

fenekku: Thank you! That is quite helpful in understanding some of the idiosyncracies of SQLAlchemy.

invenio_drafts_resources/services/draft.py
fenekku: (haha I really need to do a re-read of my code review before submitting - this was meant to be removed :sweat_smile: )

invenio_drafts_resources/services/pid_manager.py
fenekku: I think so too. Skewing towards classes is better.

PR #40, opened 2020-08-17
fix titleized name

fenekku:
just weird...


ppanero:
according to: https://github.com/inveniosoftware/cookiecutter-invenio-module/blob/master/%7B%7B%20cookiecutter.project_shortname%20%7D%7D/setup.py#L51 it should have been lowercased :/ anyway thanks for the fix :)

PR #53, opened 2020-09-18
models: integration of the latest invenio-records-resources

ppanero:
Changes:

- `drafts` -> `records` to comply with modules/hierarchy
- `DraftsBase` depends on `Record` from resources
- expires/fork should be system fields (ModelField, etc.). (example in mock module)
- Wipe create method() due to systemfields
- `drafts` revision id is called *fork_version_id* in the `models.py` and *fork_id* in the `api.py`. Should be the same: *fork_version_id*.
- There is a simple `Record` class that will later on (when services are implemented) contain a reference to its draft class (as talked IRL)
- `conceptpid` is for now a `PIDfield`. It will become a new system field `ConceptPIDField` later on when refactoring the corresponding part of the service.

Important:

- Tests were removed because they were causing a lot of import inconsistencies. 
- Tests will not pass in Travis, due again, to the import problems (removing/commenting all the code seemed extreme).
- Newly implemented tests are passing (see picture below). Are they enough?

Screenshot 2020-09-18 at 11 25 34


ppanero:
Superseeded by #54

PR #54, opened 2020-09-23
data access, service: refactor to comply with latest implementation

ppanero:


invenio_drafts_resources/services/records/service.py
```diff
@@ -0,0 +1,228 @@
...
+
+    def edit(self, id_, identity, links_config=None):
+        """Create a new revision or a draft for an existing record.
+
+        Note: Because of the workflow, this method does not accept data.
+        :param id_: record PID value.
+        """
+        self.require_permission(identity, "create")
+        # Get draft
+        try:  # Draft exists
```
zzacharo, line 120: minor: maybe just extract this logic in a `_edit_or_create` method. It will make the code easier to read :)

zzacharo (approved):

invenio_drafts_resources/services/records/service.py
ppanero: That is how it was before (what we debugged yesterday, but we need to know the `record_cls` and `draft_cls`). So:

a) We make the Draft class know which class it is drafting.
b) We can extract that function but at service level

zzacharo: Sorry I meant the `b`. Extract that function in the service level ;) I am not sure if is reusable, but I think it would make the code a bit more readable...No strong opinion though :)

ppanero: Agreed, done :)

lnielsen:
Superseded by #55

PR #59, opened 2020-09-30
links: generate draft links

fenekku:
This needs 
- marshmallow utils branch
- invenio-records-resources branch
Marking as WIP until those are merged in.


invenio_drafts_resources/resources/drafts/__init__.py
```diff
@@ -0,0 +1,23 @@
+# -*- coding: utf-8 -*-
```
lnielsen, line 1: Minor: This should be consistent with invenio-records-resources, so either 1) not move the files or 2) also perform the move in invenio-records-resources.

invenio_drafts_resources/resources/drafts/schema.py
```diff
@@ -0,0 +1,29 @@
+# -*- coding: utf-8 -*-
```
lnielsen, line 1: Minor: I would not name it ``schema.py`` but ``links.py`` instead, as this is dealing specifically with links and the file can easily be confused with schema.py for the actual data schema and the two of them shouldn't really be mixed.

tests/mock_module/resource.py
```diff
@@ -1,10 +1,21 @@
...
 
-from invenio_records_resources.resources import \
-    RecordResource as RecordResourceBase
-from invenio_records_resources.resources import \
-    RecordResourceConfig as RecordResourceConfigBase
+from marshmallow import Schema
+from marshmallow_utils.fields import Link
+from uritemplate import URITemplate
 
+# TODO: Make isort NOT want imports formatted this way
```
lnielsen, line 18: Comment: that will be a bit harder, and imply changes it 92 repositories if we change it ;-)

tests/mock_module/resource.py
```diff
@@ -17,6 +28,36 @@
...
+
+    self = Link(
+        template=URITemplate("/api/mocks/{pid_value}"),
+        permission="read",
+        params=lambda record: {'pid_value': record.pid.pid_value}
+    )
+    # TODO: Add delete, files, ...
+
+
+class DraftLinksSchema(Schema):
```
lnielsen, line 48: Comment: Seems to me that we need a small utility to write these guys:

```python
RecordLinksSchema, DraftLinksSchema = create_link_schemas('/api/mocks')
```

invenio_drafts_resources/resources/drafts/resource.py
```diff
@@ -138,10 +138,16 @@ def create(self):
...
-        # id_ and identity
-        item = cmd_func(
-            resource_requestctx.route["pid_value"],
-            g.identity
-        )
+        if cmd_name == "publish":
+            item = cmd_func(
+                resource_requestctx.route["pid_value"],
+                g.identity,
+                links_config=self.config.record_links_config
```
lnielsen, line 145: Question: Wouldn't this be potentially needed by all?

invenio_drafts_resources/resources/drafts/__init__.py
fenekku: Move also done in records-resources :heavy_check_mark:

invenio_drafts_resources/resources/drafts/schema.py
fenekku: split the difference here and named it: `schemas_links.py`

tests/mock_module/resource.py
fenekku: A man can dream... :comet: 

It's ok I will remove the comment :smile:

tests/mock_module/resource.py
fenekku: Let's wait and see... There are tons of permutation of the arguments...

invenio_drafts_resources/resources/drafts/resource.py
fenekku: A priori this is the `DraftActionResource`, so I would assume that most would return a `draft_links_config` over `record_links_config (`"publish"` action is special). But we only have 1 case, so let's wait and see...

fenekku:
Coverage decreased due to https://github.com/inveniosoftware/invenio-drafts-resources/issues/62 . Merging anyway - someone more familiar can elucidate where `is_deleted` went (it wasn't on this branch)

PR #61, opened 2020-10-02
tests: clean imports and add search test at resource level

ppanero:
this test demonstrates that https://github.com/inveniosoftware/invenio-app-rdm/issues/284 goes to invenio-records-resources level

### DIFF
--- a/tests/mock_module/api.py
+++ b/tests/mock_module/api.py
@@ -19,7 +19,7 @@ class Record(RecordBase):
     schema = ConstantField(
         '$schema', 'http://localhost/schemas/records/record-v1.0.0.json')
 
-    index = IndexField('records-record-v1.0.0', search_alias='records')
+    index = IndexField('records-record-v1.0.0', search_alias='draftsresources-records')
 
 
 class Draft(DraftBase):
@@ -32,4 +32,4 @@ class Draft(DraftBase):
     schema = ConstantField(
         '$schema', 'http://localhost/schemas/records/record-v1.0.0.json')
 
-    index = IndexField('drafts-draft-v1.0.0', search_alias='drafts')
+    index = IndexField('drafts-draft-v1.0.0', search_alias='draftsresources-drafts')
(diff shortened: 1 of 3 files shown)


tests/mock_module/api.py
```diff
@@ -19,7 +19,7 @@ class Record(RecordBase):
     schema = ConstantField(
         '$schema', 'http://localhost/schemas/records/record-v1.0.0.json')
 
-    index = IndexField('records-record-v1.0.0', search_alias='records')
+    index = IndexField('records-record-v1.0.0', search_alias='draftsresources-records')
```
lnielsen, line 22: Question: what's the reason for this change?

ppanero: @lnielsen The agreement we got to have the folders like:

```
draftsresources/    # equivalent to rdmrecords
    records/
        ...
    drafts/
        ...
```

To allow index/alias creation. How `invenio-search` works is that it will create:
- `draftsresources`: queries all indices
- `draftsresources -records`: queries all indices inside that folder
- `draftsresources -drafts`: same as above but for drafts

meaning it prefixes all sort of in a recursive way.

lnielsen: Ahh, got it :-) Slow today.... 😄

tests/resources/test_record_resource.py
```diff
@@ -168,6 +168,23 @@ def test_publish_draft(client, input_data):
...
 
+def test_search_draft(client, input_data):
+    response = client.post(
+        "/mocks", data=json.dumps(input_data), headers=HEADERS)
+    recid = response.json['id']
+
+    assert response.status_code == 201
+    _assert_single_item_response(response)
+
+    response = client.get("/mocks?q=", headers=HEADERS)
```
fenekku, line 179: This test is failing because we are hitting the published record endpoints and there are no published records; there is only a draft at this point. Publishing and searching is a test case already covered in invenio-records-resources, so I wouldn't worry about adding this test here. The `q=` case is (or should be) in invenio-records-resources I think.

ppanero:
Agreeing with @fenekku and taking into account said test is already in invenio-records-resources I am closing this PR and porting the search alias fix to https://github.com/inveniosoftware/invenio-drafts-resources/pull/63

PR #65, opened 2020-10-08
tests: unskip is_delete tests and bump invenio-records

ppanero:


ppanero:
Closed in favour of #66

PR #82, opened 2020-12-07
services: relax save conditions but validate when publishing

### LINKED ISSUE #81: Lax save and strict publish
The RecordDraftService must be able to save
- completely empty drafts
- drafts partially (but correctly) filled

(even missing required fields in both cases)

The RecordDraftService must not save and must generate validation errors
- filled fields that are incorrect

The RecordDraftService must not publish and must generate validation errors
- missing required fields
- (incorrectly filled fields should have been caught by above but it's fine to check them again)

The goal:

As a depositor, I want to fill out my draft as much or as little as I want, but be prevented from publishing if I am missing required fields in order to have a flexible but strict-where-it-matters workflow.

fenekku:
- closes #81 
- :warning: requires https://github.com/inveniosoftware/invenio-records-resources/pull/173
- starts December release line: 0.7.x

### DIFF
--- a/invenio_drafts_resources/services/records/service.py
+++ b/invenio_drafts_resources/services/records/service.py
@@ -204,9 +204,6 @@ class RecordDraftService(RecordService):
         # Get draft
         draft = self.draft_cls.pid.resolve(id_, registered_only=False)
 
-        # Fully validate draft now
-        # TODO: Add error_map for `ValidationError` in the resource config
-
         data, _ = self.schema.load(
             identity,
             data=draft.dumps(),
--- a/invenio_drafts_resources/services/records/service.py
+++ b/invenio_drafts_resources/services/records/service.py
@@ -206,8 +206,15 @@ class RecordDraftService(RecordService):
 
         # Fully validate draft now
         # TODO: Add error_map for `ValidationError` in the resource config
-        # FIXME: Check partial vs absolute validation
-        # data, _ = self.schema.load(identity, data=draft.dumps())
+
+        data, _ = self.schema.load(
+            identity,
+            data=draft.dumps(),
+            pid=draft.pid,
+            record=draft,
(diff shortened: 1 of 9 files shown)


invenio_drafts_resources/services/records/service.py
```diff
@@ -190,8 +206,15 @@ def publish(self, id_, identity, links_config=None):
 
         # Fully validate draft now
         # TODO: Add error_map for `ValidationError` in the resource config
```
ppanero, line 208: minor: wondering if this old TODO makes sense now, as far as I understand errors are handled but not as an exception.

tests/services/test_record_service.py
```diff
@@ -56,6 +48,26 @@ def test_create_draft(app, service, identity_simple, input_data):
...
+    }
+
+    draft = service.create(identity_simple, input_data)
+    draft_dict = draft.to_dict()
+
+    assert draft['id']
+    assert draft['conceptid']
+    assert draft._record.pid.status == PIDStatus.NEW
+    assert draft._record.conceptpid.status == PIDStatus.NEW
+    assert draft_dict['errors']['metadata']['title']
```
ppanero, line 68: minor: what do you think about putting this in a different test, to make things more explicit.

ppanero (commented):
LGTM, same as for the resource PR :)

lnielsen (approved):

slint (approved):
LGTM, I think also exactly what I had in mind at some point for the loader :)

invenio_drafts_resources/services/records/service.py
fenekku: Right. I will remove those.

tests/services/test_record_service.py
fenekku: Added the check for no errors in the happy path create draft test instead.

PR #85, opened 2020-12-10
records: add draft pid provider

### LINKED ISSUE #84: PID status overwritten for `RecordIdProviderV2`
Package version (if known): 0.7.0

## Describe the bug
https://github.com/inveniosoftware/invenio-drafts-resources/blob/master/invenio_drafts_resources/records/api.py#L19
This line causes all the other record types using `RecordIdProviderV2` to be created with `PIDStatus.NEW` instead of registered, breaking any resource details view in the REST API.

## Steps to Reproduce
1. install ay module using RecordIdProviderV2 (f.e. invenio-vocabularies)
2. Run invenio vocabularies import languages
3. Try to access /api/vocabularies/languages/<pid>

## Expected behavior

Pids should be registered with status `REGISTERED` if not a draft record

kprzerwa:
closes #84

### DIFF
--- a/invenio_drafts_resources/records/api.py
+++ b/invenio_drafts_resources/records/api.py
@@ -16,7 +16,11 @@ from invenio_records_resources.records.systemfields import PIDField, \
     PIDStatusCheckField
 from sqlalchemy.orm.exc import NoResultFound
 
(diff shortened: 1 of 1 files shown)


slint (approved):
LGTM

PR #96, opened 2021-02-23
service: pass default_filter for records and drafts search

zzacharo:
closes https://github.com/inveniosoftware/invenio-app-rdm/issues/563

### DIFF
--- a/invenio_drafts_resources/services/records/service.py
+++ b/invenio_drafts_resources/services/records/service.py
@@ -9,6 +9,7 @@
 
 """RecordDraft Service API."""
 
+from elasticsearch_dsl.query import Q
 from invenio_db import db
 from invenio_records_resources.services import RecordService
 from sqlalchemy.orm.exc import NoResultFound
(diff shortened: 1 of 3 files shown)


invenio_drafts_resources/services/records/service.py
```diff
@@ -65,6 +66,7 @@ def search_drafts(self, identity, params=None, links_config=None,
             links_config=links_config,
             es_preference=es_preference,
             record_cls=self.draft_cls,
+            extra_filter=Q('term', **{'has_draft': False})
```
lnielsen, line 69: Comment: You can do ``Q('term', has_draft=False)``. The other syntax is just necessary when you have e.g. dots.

lnielsen, line 69: Question: It's not ``has_draft==True`` we want?

lnielsen (approved):

invenio_drafts_resources/services/records/service.py
zzacharo: No because `has_draft` is set to False for the published records and the drafts....True is only for the published records with an active draft...but these records we get from the drafts index and we don't want to include them in the published ones...

PR #107, opened 2021-03-10
records: initial versioning support in data layer

lnielsen:
⚠️ Depends on https://github.com/inveniosoftware/invenio-records/pull/249

closes inveniosoftware/invenio-rdm-records#104

### DIFF
--- a/invenio_drafts_resources/records/api.py
+++ b/invenio_drafts_resources/records/api.py
@@ -23,6 +23,15 @@ class DraftRecordIdProviderV2(RecordIdProviderV2):
     default_status_with_obj = PIDStatus.NEW
 
 
+class ParentRecord(RecordBase):
+    """Parent record API."""
+
+    # Configuration
+    model_cls = None
+
+    pid = PIDField('id', provider=DraftRecordIdProviderV2, delete=True)
+
+
 class Record(RecordBase):
     """Record base API.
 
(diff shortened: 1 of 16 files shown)


invenio_drafts_resources/records/api.py
```diff
@@ -23,6 +23,15 @@ class DraftRecordIdProviderV2(RecordIdProviderV2):
     default_status_with_obj = PIDStatus.NEW
 
 
+class ParentRecord(RecordBase):
+    """Parent record API."""
+
+    # Configuration
+    model_cls = None
+
+    pid = PIDField('id', provider=DraftRecordIdProviderV2, delete=True)
```
lnielsen, line 32: We moving the concept pid to here.

invenio_drafts_resources/records/api.py
```diff
@@ -32,39 +41,52 @@ class Record(RecordBase):
...
-            # New version
-            record = cls.get_record(draft.id)
-        except NoResultFound:
-            # New revision
-            record = cls.create(
-                {}, id_=draft.id, pid=draft.pid, conceptpid=draft.conceptpid)
-
-        # NOTE: Merge pid/conceptpid into the current db session if not already
-        # in the session.
+    def create_from(cls, draft):
```
lnielsen, line 59: Not pretty - it would be better to have something on the system fields, but the hooks are not easily extended with new ones in Invenio-Records.

invenio_drafts_resources/records/systemfields/parent.py
```diff
@@ -0,0 +1,97 @@
...
+        if self._hard_delete or self._soft_delete:
+            self.delete(record, force=force)
+
+    def pre_dump(self, record, data, dumper=None):
+        """Called before a record is dumped."""
+        parent = getattr(record, self.attr_name)
+        # Use the default dumper for the parent (since the records/drafts
+        # dumpers won't work on a parent)
+        if parent:
+            data[self.attr_name] = parent.dumps(dumper=None)
```
lnielsen, line 75: A bit tricky here......you could potentially use a different dumper (say file dumper instead of ES dumper), and then the parent woudl always use the ES dumper...

tests/mock_module/api.py
```diff
@@ -24,6 +38,13 @@ class Record(RecordBase):
         search_alias='draftsresources-records'
     )
 
+    parent = ParentField(
```
lnielsen, line 41: I still have to updated both the records and drafts schema to v2.0.0

invenio_drafts_resources/records/systemfields/parent.py
max-moser: hmm... have i understood correctly that the main problem here is, that the records' dumpers won't work on their parents because the structure is fundamentally different between records and parents?

would some kind of mapping function between the record and parent dumpers help here?
```python
def translate_record_to_parent_dumper(record_dumper):
  if isinstance(record_dumper, ESDumper):
    return ParentESDumper()
  elif:
    # ...
```

max-moser (approved):
LGTM, added some comments/questions/minor remarks

invenio_drafts_resources/records/systemfields/parent.py
lnielsen: > the records' dumpers won't work on their parents because the structure is fundamentally different between records and parents?

Yep correct

> would some kind of mapping function between the record and parent dumpers help here?

yep something along the lines.... it ties a bit into how you could potentially override the indexing to tweak it to your needs since that's usually via a dumper.....on the other hand, we likely don't allow people to stick stuff in the parent record

invenio_drafts_resources/records/models.py
```diff
@@ -13,6 +13,24 @@
...
+        @declared_attr
+        def parent_id(cls):
+            return db.Column(UUIDType, db.ForeignKey(parent_record_cls.id))
+
+        @declared_attr
+        def parent(cls):
+            return db.relationship(parent_record_cls)
+
+        # TODO: Add parent_order and/or parent_latest
+        # TODO: Should both records and drafts have an order?
```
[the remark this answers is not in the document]
lnielsen: Yes, that's my current thinking so you can do something like ``select * from rdm_records_metadata where parent_id=... order by parent_order desc``

PR #110, opened 2021-03-12
records/service: implement new version creation

lnielsen:
closes inveniosoftware/invenio-rdm-records#435

This is the Draft-Resources part of versioning. The RDM-Records part is still missing, but this one can be reviewed independently from RDM-Records.

### DIFF
--- a/invenio_drafts_resources/records/api.py
+++ b/invenio_drafts_resources/records/api.py
@@ -1,12 +1,26 @@
 # -*- coding: utf-8 -*-
 #
-# Copyright (C) 2020 CERN.
+# Copyright (C) 2020-2021 CERN.
 #
 # Invenio-Drafts-Resources is free software; you can redistribute it and/or
 # modify it under the terms of the MIT License; see LICENSE file for more
 # details.
 
-"""Draft API."""
+"""Record, Draft and Parent Record API classes.
+
+These classes belongs to the  data access layer and MUST ONLY be accessed from
+within the service layer. It's wrong to use these classes in the presentation
+layer.
+
+A record and a draft share a single parent record. The parent record is used
+to store properties common to all versions of a record (e.g. access control).
+
+The draft and record share the same UUID, and thus both also share a single
+persistent identifier. The parent record has its own UUID and own persistent
+identifier.
+"""
+
+import uuid
 
 from invenio_pidstore.models import PIDStatus
 from invenio_pidstore.providers.recordid_v2 import RecordIdProviderV2
@@ -16,13 +30,21 @@ from invenio_records_resources.records.systemfields import PIDField, \
     PIDStatusCheckField
 from sqlalchemy.orm.exc import NoResultFound
 
+from .systemfields import ParentField, VersionsField
 
+
+#
+# Persistent identifier providers
+#
 class DraftRecordIdProviderV2(RecordIdProviderV2):
     """Draft PID provider."""
 
     default_status_with_obj = PIDStatus.NEW
 
 
+#
+# Record API classes
+#
 class ParentRecord(RecordBase):
     """Parent record API."""
 
@@ -33,57 +55,66 @@ class ParentRecord(RecordBase):
 
 
 class Record(RecordBase):
-    """Record base API.
+    """Record API."""
 
-    Note: This class is meant to work along a with a draft class.
-    """
+    #
+    # Configuration to be set by a subclass
+    #
 
-    # Configuration
+    #: The record's SQLAlchemy model class. Must be set by the subclass.
     model_cls = None
+    #: The parent state's SQLAlchemy model class. Must be set by the subclass.
+    versions_model_cls = None
+    #: The parent record's API class. Must be set by the subclass.
+    parent_record_cls = None
 
+    #
+    # System fields
+    #
+    #: The internal persistent identifier. Records and drafts share UUID.
     pid = PIDField('id', provider=DraftRecordIdProviderV2, delete=True)
 
+    #: System field to check if a record has been published.
     is_published = PIDStatusCheckField(status=PIDStatus.REGISTERED)
 
-    #: List of field names (strings) to copy from the draft on create.
-    create_from_draft = []
+    #: The parent record - the draft is responsible for creating the parent.
+    parent = ParentField(
+        ParentRecord, create=False, soft_delete=False, hard_delete=False)
 
-    #: List of field names (strings) to copy from the draft on update.
-    update_from_draft = []
+    #: Version relationship
+    versions = VersionsField(create=True, set_latest=True)
 
-    # TODO: Below three methods create_from(), update_from() and register() has
-    # to be refactored. They are accessing "parent" but it may not be defined.
-    # Instead, this work should be delegated to the system fields, but it's not
-    # easy to add a new pre/post_create_from pre/post_register hook.
     @classmethod
-    def create_from(cls, draft):
-        """Create a new record from a draft."""
-        record = cls.create(
-            {},
-            id_=draft.id,
-            pid=draft.pid,
-            **{f: getattr(draft, f) for f in cls.create_from_draft}
-        )
+    def publish(cls, draft):
+        """Publish a draft as a new record.
 
-        # NOTE: Merge pid into the current db session if not already in the
-        # session.
-        cls.pid.session_merge(record)
-        record.parent.__class__.pid.session_merge(record.parent)
+        If a record already exists, we simply get the record. If a draft has
+        not yet been published, we create the record.
 
-        record.update_from(draft)
-
+        The caller is responsible for registering the internal persistent
+        identifiers (see ``register()``).
+        """
+        if draft.is_published:
+            record = cls.get_record(draft.id)
+        else:
+            record = cls.create(
+                {},
+                # A draft and record share UUID, so we reuse the draft's id/pid
+                id_=draft.id,
+                pid=draft.pid,
+                # Link the record with the parent record and set the versioning
+                # relationship.
+                parent=draft.parent,
+                versions=draft.versions,
+            )
+            # Merge the PIDs into the current db session if not already in the
+            # session.
+            cls.pid.session_merge(record)
+            cls.parent_record_cls.pid.session_merge(record.parent)
         return record
 
-    def update_from(self, draft):
-        """Update a record from a draft."""
-        # Overwrite data
-        self.update(**draft)
-        for f in self.update_from_draft:
-            setattr(self, f, getattr(draft, f))
-        return self
-
     def register(self):
-        """Register the persistent identifiers associated with the record."""
+        """Register the internal persistent identifiers."""
         if not self.parent.pid.is_registered():
             self.parent.pid.register()
             self.parent.commit()
@@ -93,12 +124,90 @@ class Record(RecordBase):
 class Draft(Record):
     """Draft base API for metadata creation and manipulation."""
 
-    # WHY: We want to force the model_cls to be specified by the user
-    # No default one is given, only the base.
-    model_cls = None
+    #
+    # Configuration to be set by a subclass
+    #
 
+    #: The record's SQLAlchemy model class. Must be set by the subclass.
+    model_cls = None
+    #: The parent state's SQLAlchemy model class. Must be set by the subclass.
+    versions_model_cls = None
+    #: The parent record's API class. Must be set by the subclass.
+    parent_record_cls = None
+
+    #
+    # System fields
+    #
+
+    #: The internal persistent identifier. Records and drafts share UUID.
     pid = PIDField('id', provider=DraftRecordIdProviderV2, delete=False)
 
+    #: The parent record - the draft is responsible for creating the parent.
+    parent = ParentField(
+        ParentRecord, create=True, soft_delete=False, hard_delete=True)
+
+    #: Version relationship
+    versions = VersionsField(create=True, set_next=True)
+
+    #: The expiry date of the draft.
     expires_at = ModelField()
 
+    #: Revision id of record from which this draft was created.
     fork_version_id = ModelField()
+
+    @classmethod
+    def new_version(cls, record):
+        """Create a draft for a new version of a record.
+
+        The caller is responsible for:
+        1) checking if a draft for a new version already exists
+        2) moving the record data into the draft data.
+        """
+        return cls.create(
+            {},
+            # We create a new id, because this is for a new version.
+            id=uuid.uuid4(),
+            # Links the draft with the same parent (i.e. a new version).
+            parent=record.parent,
+            versions=record.versions,
+            # New drafts without a record (i.e. unpublished drafts) must set
+            # the fork version id to None.
+            fork_version_id=None,
+        )
+
+    @classmethod
+    def edit(cls, record):
+        """Create a draft for editing an existing version of a record."""
+        try:
+            # We soft-delete a draft once it has been published, in order to
+            # keep the version_id counter around for optimistic concurrency
+            # control (both for ES indexing and for REST API clients)
+            draft = cls.get_record(record.id, with_deleted=True)
+            if draft.is_deleted:
+                draft.undelete()
+                # Below line is needed to dump PID back into the draft.
+                draft.pid = record.pid
+                # Ensure record is link with the parent
+                draft.parent = record.parent
+                draft.versions = record.versions
+                # Ensure we record the revision id we forked from
+                draft.fork_version_id = record.revision_id
+                # Note, other values like bucket_id values was kept in the
+                # soft-deleted record, so we are not setting them again here.
+        except NoResultFound:
+            # If a draft was ever force deleted, then we will create the draft.
+            # This is a very exceptional case as normally, when we edit a
+            # record then the soft-deleted draft exists and we are in above
+            # case.
+            draft = cls.create(
+                {},
+                # A draft to edit a record must share the id and uuid.
+                id_=record.id,
+                pid=record.pid,
+                # Link it with the same parent record
+                parent=record.parent,
+                versions=record.versions,
+                # Record which record version we forked from.
+                fork_version_id=record.revision_id,
+            )
+        return draft
--- a/invenio_drafts_resources/records/models.py
+++ b/invenio_drafts_resources/records/models.py
@@ -17,24 +17,109 @@ from sqlalchemy.ext.declarative import declared_attr
 from sqlalchemy_utils.types import UUIDType
 
 
-def ParentRecordMixin(parent_record_cls):
+class ParentRecordMixin:
     """A mixin factory that add the foreign keys to the parent record.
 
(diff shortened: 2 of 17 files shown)


invenio_drafts_resources/records/api.py
```diff
@@ -33,57 +55,66 @@ class ParentRecord(RecordBase):
...
-        )
-
-        # NOTE: Merge pid into the current db session if not already in the
-        # session.
-        cls.pid.session_merge(record)
-        record.parent.__class__.pid.session_merge(record.parent)
-
-        record.update_from(draft)
-
+    def publish(cls, draft):
```
lnielsen, line 88: Comment: I've moved this logic from ``service.publish()`` into here to ensure that it's clearly defined in the API class how a draft is published. This is because if e.g. ``parent`` or ``versions`` is left-out of the ``create()`` the versioning get's messed up.

invenio_drafts_resources/records/api.py
```diff
@@ -33,57 +55,66 @@ class ParentRecord(RecordBase):
...
+                parent=draft.parent,
+                versions=draft.versions,
+            )
+            # Merge the PIDs into the current db session if not already in the
+            # session.
+            cls.pid.session_merge(record)
+            cls.parent_record_cls.pid.session_merge(record.parent)
         return record
 
-    def update_from(self, draft):
```
lnielsen, line 77: Comment: All updates of data is now the responsibility of a component (see e.g. MetadataComponent). Doing record.update(**draft) is very unreliable and bypasses all the system fields.

invenio_drafts_resources/records/api.py
```diff
@@ -93,12 +124,90 @@ def register(self):
...
+    versions = VersionsField(create=True, set_next=True)
+
+    #: The expiry date of the draft.
     expires_at = ModelField()
 
+    #: Revision id of record from which this draft was created.
     fork_version_id = ModelField()
+
+    @classmethod
+    def new_version(cls, record):
```
lnielsen, line 159: Comment: Same reasoning here as with ``publish()`` above.

invenio_drafts_resources/records/api.py
```diff
@@ -93,12 +124,90 @@ def register(self):
...
+            # Links the draft with the same parent (i.e. a new version).
+            parent=record.parent,
+            versions=record.versions,
+            # New drafts without a record (i.e. unpublished drafts) must set
+            # the fork version id to None.
+            fork_version_id=None,
+        )
+
+    @classmethod
+    def edit(cls, record):
```
lnielsen, line 179: Comment: Same reasoning here as with ``publish()`` above.

invenio_drafts_resources/records/models.py
```diff
@@ -17,24 +17,109 @@
...
     """A mixin factory that add the foreign keys to the parent record.
 
-    It is intended to be added to the "child" record class, e.g.:
-    ``class MyRecord(RecordBase, ParentRecordMixin(MyRecordParentClass))``.
+    Usage:
+
+    .. code-block:: python
+
+        class MyRecord(db.Model, RecordMetadataBase, ParentRecordMixin):
+            parent_record_model = ParentRecordMetadata
```
lnielsen, line 28:
```suggestion
            __parent_record_model__ = ParentRecordMetadata
```

invenio_drafts_resources/records/models.py
```diff
@@ -17,24 +17,109 @@
...
+        # exists via database-level trigger.
+        return db.Column(UUIDType, db.ForeignKey(
+            cls.__parent_record_model__.id, ondelete="RESTRICT"))
+
+    @declared_attr
+    def parent(cls):
+        """Relationship to parent record."""
+        return db.relationship(cls.__parent_record_model__)
+
+    parent_index = db.Column(db.Integer, nullable=True)
```
lnielsen, line 46: Comment: after some discussions with Alex/Nico/Jose, then it seemd ``parent_index`` was the best name.

invenio_drafts_resources/records/models.py
```diff
@@ -17,24 +17,109 @@
 from sqlalchemy_utils.types import UUIDType
 
 
-def ParentRecordMixin(parent_record_cls):
+class ParentRecordMixin:
```
lnielsen, line 20: Comment: Refactored to use proper SQLAlchemy base classes.

invenio_drafts_resources/records/models.py
```diff
@@ -17,24 +17,109 @@
...
-        @declared_attr
-        def parent(cls):
-            return db.relationship(parent_record_cls)
-
-        # TODO: Add parent_order and/or parent_latest
-        # TODO: Should both records and drafts have an order?
-    return Mixin
+
+    @declared_attr
+    def next_id(cls):
```
lnielsen, line 108: ``next_id`` -> ``next_draft_id``

invenio_drafts_resources/records/systemfields/parent.py
```diff
@@ -33,23 +32,49 @@ def __init__(self, parent_record_cls, key=None, create=False,
...
+        # The logic implemented here, thus on a hard delete tries to delete the
+        # the parent record and relies on the database integrity constraints
+        # to prevent the parent record from being deleted in case more
+        # drafts/records exists.
         parent = getattr(record, self.attr_name)
         if parent:
             if force and self._hard_delete:
-                parent.delete(force=True)
+                try:
+                    with db.session.begin_nested():
```
lnielsen, line 72: Comment: If an exception is detected, we rollback to the save point (otherwise the entire transaction needs to be rolled back). See also https://invenio-db.readthedocs.io/en/latest/session_management.html#what-happens-with-exceptions-in-nested-transactions

invenio_drafts_resources/records/systemfields/versions.py
```diff
@@ -0,0 +1,259 @@
...
+
+
+def uuid_or_none(val):
+    """Convert string to UUID object."""
+    if val is not None and not isinstance(val, uuid.UUID):
+        return uuid.UUID(val)
+    return val
+
+
+class VersionsManager:
```
lnielsen, line 28: Comment: Later we could add more methods in order to query for the other records via the database etc.

invenio_drafts_resources/records/api.py
```diff
@@ -93,12 +124,90 @@ def register(self):
...
 
+    #: Revision id of record from which this draft was created.
     fork_version_id = ModelField()
+
+    @classmethod
+    def new_version(cls, record):
+        """Create a draft for a new version of a record.
+
+        The caller is responsible for:
+        1) checking if a draft for a new version already exists
```
fenekku, line 163: Should we explicitly address our discussion about creating a new version only for most recent version record *for now* (in the future we want to allow the version switching) here?

(Probably technical hurdles I am not seeing: Would it make sense to (eventually perhaps) have the same mechanism as for above `draft.is_published` but for latest version here and so have 1) be taken care of. This would be consistent with what `publish` does and what `edit` does. Don't know if that is possible yet though.)

fenekku (commented):
I have to go to lunch so here are preliminary comments. I will continue after lunch.

The `private` ones are for me: questions I have that might be resolved by reading the remaining code files.

invenio_drafts_resources/records/api.py
```diff
@@ -33,57 +55,66 @@ class ParentRecord(RecordBase):
...
+    """Record API."""
 
-    Note: This class is meant to work along a with a draft class.
-    """
+    #
+    # Configuration to be set by a subclass
+    #
 
-    # Configuration
+    #: The record's SQLAlchemy model class. Must be set by the subclass.
```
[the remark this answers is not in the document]
lnielsen: Yes. It's inline sphinx autodoc syntax. It's equivalent to:

```
model_cls = None
"""...."""
```

invenio_drafts_resources/records/api.py
```diff
@@ -33,57 +55,66 @@ class ParentRecord(RecordBase):
...
     pid = PIDField('id', provider=DraftRecordIdProviderV2, delete=True)
 
+    #: System field to check if a record has been published.
     is_published = PIDStatusCheckField(status=PIDStatus.REGISTERED)
 
-    #: List of field names (strings) to copy from the draft on create.
-    create_from_draft = []
+    #: The parent record - the draft is responsible for creating the parent.
+    parent = ParentField(
+        ParentRecord, create=False, soft_delete=False, hard_delete=False)
```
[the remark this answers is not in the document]
lnielsen: The value is just used in a single place to do a ``isinstance(val, ParentRecord)`` so in this case it's ok as long as the base classes that's passed in. Its because I'm relying on ``RelatedModelField`` and this value is not that easy to overwrite.

invenio_drafts_resources/records/api.py
lnielsen: Yeah, I was not sure what to do here.

For reference here's the part of the service code:
```
# Draft for new version already exists? if so return it
if record.versions.next_draft_id:
    next_draft = self.draft_cls.get_record(
        record.versions.next_draft_id)
    return self.result_item(
        self, identity, next_draft, links_config=links_config)

# Draft for new version does not exists, so create it
next_draft = self.draft_cls.new_version(record)

# Get the latest published record if it's not the current one.
if not record.versions.is_latest:
    record = self.record_cls.get_record(record.versions.latest_id)
```

The tricky part here is to know that the draft already exists, and return it immediately (this part of the code aligns with ``service.edit()``.

Now seeing it again, I honestly think the method works with any record (not only the latest), so probably the comment is actually not relevant.

invenio_drafts_resources/records/api.py
```diff
@@ -33,57 +55,66 @@ class ParentRecord(RecordBase):
...
+    #
+    # Configuration to be set by a subclass
+    #
 
-    # Configuration
+    #: The record's SQLAlchemy model class. Must be set by the subclass.
     model_cls = None
-
+    #: The parent state's SQLAlchemy model class. Must be set by the subclass.
+    versions_model_cls = None
```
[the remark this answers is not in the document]
lnielsen: It's this one: https://github.com/inveniosoftware/invenio-drafts-resources/pull/110/files/8d166da87291500e66a89ff3fab723cf69f50e1f#diff-b32aa79b3a09b3864fa105c1bc774cfd8d0646c1e5f3e3d2a6f9c58455ece88dR50

invenio_drafts_resources/records/api.py
```diff
@@ -33,57 +55,66 @@ class ParentRecord(RecordBase):
...
-    #: List of field names (strings) to copy from the draft on create.
-    create_from_draft = []
+    #: The parent record - the draft is responsible for creating the parent.
+    parent = ParentField(
+        ParentRecord, create=False, soft_delete=False, hard_delete=False)
 
-    #: List of field names (strings) to copy from the draft on update.
-    update_from_draft = []
+    #: Version relationship
+    versions = VersionsField(create=True, set_latest=True)
```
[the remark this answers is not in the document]
lnielsen: Regarding naming my thinking here was, that we'll have a REST API serialization looking similar to this:

```python
{
    "parent": {
        "id": "",
        "access": ...
    },
    "versions": {
        "is_latest": true,
        "index": 1,
        ...
    }
}
```

Parent thus dealing mainly with common metadata for all records, and versions dealing with if you're he latest or not, which index etc.

tests/records/test_api.py
```diff
@@ -47,12 +49,141 @@ def test_draft_create_parent(app, db):
     assert draft.pid.object_uuid != draft.parent.pid.object_uuid
 
 
+def test_draft_create_parent_state(app, db):
+    """Test draft creation of the parent record."""
+    draft = Draft.create({})
+    db.session.commit()
+
+    # Assert that associated objects where created
```
fenekku, line 57:
```suggestion
    # Assert that associated objects were created
```

fenekku (approved):
Most of my comments are about clarifications. 

I am now realizing that these systemfields are really used as a namespace over sets of methods that provide a pythonic API / cache to the underlying model. Long chaining can be brittle however (`record.model.parent_index = self.parent_index`) and some are non-trivial to grasp
(`record.versions.is_latest_draft` - is a record a draft? a version a draft?)

invenio_drafts_resources/records/models.py
lnielsen: What about just ``index``?

invenio_drafts_resources/records/api.py
```diff
@@ -93,12 +124,90 @@ def register(self):
...
+    @classmethod
+    def edit(cls, record):
+        """Create a draft for editing an existing version of a record."""
+        try:
+            # We soft-delete a draft once it has been published, in order to
+            # keep the version_id counter around for optimistic concurrency
+            # control (both for ES indexing and for REST API clients)
+            draft = cls.get_record(record.id, with_deleted=True)
+            if draft.is_deleted:
+                draft.undelete()
```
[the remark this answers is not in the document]
lnielsen: We issue a soft-delete (i.e. draft stays in the table), and then issue a delete to Elasticsearch. Elasticsearch then keeps around the UUID + version id of the deleted record for 1 min. That means, that if in that 1 min, somebody goes and press the edit button again (i.e. we're creating a draft), we need to index the same UUID again. If the version id we send is lower than the one we used for the ES delete operation (e.g. in case we hard-deleted the record), ES will refuse the operation because it thinks it's an old operation that arrived late. That's why we need to keep the draft version id around, so that we send an ES index operation with a version id that's higher than the version id we used in the ES delete.......the way we keep the version_id around is by simply soft-deleting the draft, but we coudl also just have stored the version id somewhere else.

invenio_drafts_resources/records/models.py
```diff
@@ -17,24 +17,109 @@
...
+    """The version index of the record."""
+
+
+class ParentRecordStateMixin:
+    """Database model mixin to keep the state of the latest and next version.
+
+    We keep this data outside the parent record itself, because we want to
+    update it without impacting the parent record's version counter. The
+    version counter in the parent record we use to determine if we have to
+    reindex all record versions.
```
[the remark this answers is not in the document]
lnielsen: We need to reindex all when the parent record changes (e.g. access control changes). We do not need to reindex all in the case where is_latest/is_latest_draft doesn't change.

invenio_drafts_resources/records/models.py
```diff
@@ -17,24 +17,109 @@
...
+        If no record has been published, the value is None.
+        """
+        return db.Column(
+            UUIDType,
+            db.ForeignKey(cls.__record_model__.id),
+            nullable=True,
+        )
+
+    latest_index = db.Column(db.Integer, nullable=True)
+    """The index of the latest published record.
```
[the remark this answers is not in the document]
lnielsen: The reasoning for both ``latest_id`` and ``latest_index`` is that we want to avoid expensive select queries. Both properties could be queried by a select similar to:

```sql
SELECT id as latest_id, parent_index as latest_index FROM rdm_records_metadata WHERE parent_id=:id ORDER BY parent_index DESC LIMIT 1
```

Instead, we can do a single primary key look like this:

```sql
SELECT latest_id, latest_index FROM rdm_versions_state WHERE parent_id=:id
```

It's a classic search problem - you can either have 1) fast inserts and expensive selects OR 2) expensive inserts and fast selects :-)

invenio_drafts_resources/records/systemfields/versions.py
```diff
@@ -0,0 +1,259 @@
...
+        """Check if the record/draft id is the latest published record id."""
+        return self.latest_id == self._record.id
+
+    @property
+    def is_latest_draft(self):
+        """Check if the record/draft id is the latest draft id."""
+        if self.next_draft_id:
+            return self.next_draft_id == self._record.id
+        else:
+            return self.latest_id == self._record.id
```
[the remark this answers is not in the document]
lnielsen: It's related to the list uploads - i.e. when we search drafts we search over both records and drafts. 

The list uploads page can show only the latest version. This means that we apply a filter that includes only records/drafts that have ``record.is_latest_draft == True``. 

There's 3 cases where above condition has to return True:

1) The latest published record has only a record (i.e. it's not being edited) - we display the record
2) The latest published record has a record and draft (i.e. it's being edited) - we display the draft
3) A next draft exists (i.e. the we created a new version, but didn't publish yet) - we display the next draft.

The if-part is related to (3), then else-part is related to (1) and (2).

invenio_drafts_resources/records/models.py
fenekku: Some kind of indication that this is for versioning would be nicer or if a one word is much preferred, then `order` or `rank` would be better is my take :)

invenio_drafts_resources/records/api.py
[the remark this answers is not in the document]
fenekku: If setting `parent_record_cls` doesn't affect the `ParentField` then, should we just not pass `ParentRecord` as an argument to `ParentField` ? ... we can discuss this one

invenio_drafts_resources/records/api.py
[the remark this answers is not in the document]
fenekku: Ah, got it. So ES has its own "versioning" mechanism that it uses to determine that insertion of a deleted document with a more recent "version id" is fine. (It's not that we are telling ES to cancel deletion. We just tell it to insert one with higher version and that is fine whether the 1min mark has elapsed or not).

invenio_drafts_resources/records/models.py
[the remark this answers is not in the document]
fenekku: Got it: in our case we are making the insert a bit more involved (need to update `latest_index` / `latest_id` in this table in addition to the regular `rdm_records_metadata` table), but we get a faster select (which is used in insertion ;) )

invenio_drafts_resources/records/systemfields/versions.py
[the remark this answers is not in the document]
fenekku: So list uploads when "all versions" toggle switch is off shows:

```
Draft A (draft of an unpublished yet record)
Record B (published record with no drafts)
Draft C (draft of a published record)
Draft D2 (draft of the new -2nd- version of record D)
```

When toggle switch is on:

```
Draft A (draft of an unpublished yet record)
Record B (published record with no drafts)
Draft C (draft of a published record)
Draft D2 (draft of the new -2nd- version of record D)
Record D (published record for first version of record D) 
```
?


Like in overall review message, what is not clear to me is whether `record.versions.is_latest_draft` means:

a) the latest version of this record is a draft 
b) the current record is the latest version and there is a draft of it

I think it's a) (really I think it's a c) I don't grasp yet :smile: ). Naming could help:

a) -> `record.versions.latest_is_draft` (read as: the record's latest version is draft)
b) -> `record.versions.has_latest_draft` (read as: the record has the latest draft)

(same naming difficulty because is `versions` the subject of these methods or is it just a namespace? Plus record is overloaded, but that's another can of worms :sweat_smile: )

PR #111, opened 2021-03-15
resources: add records versions resource

zzacharo:


invenio_drafts_resources/resources/records/config.py
```diff
@@ -21,3 +24,19 @@ class RecordResourceConfig(_RecordResourceConfig):
     draft_links_config = {
         "record": DraftLinksSchema
     }
+
+class RecordVersionsResourceConfig(_RecordResourceConfig):
+    """Record resource version config."""
+
+    list_route = "/records/<pid_value>/versions"
+
+    item_route = "/records/<pid_value>/versions/latest"
```
zzacharo, line 33: Because we inherit from `CollectionResource` we need to define `item_route` or override the `create_url_rules` function and return only the list_route. But, I thought that the item_route can be used for displaying the latest version, so we don't have to create yet another resource...

invenio_drafts_resources/resources/records/resource.py
```diff
@@ -28,3 +34,49 @@ def create(self):
...
+
+    default_config = RecordVersionsResourceConfig
+
+    def __init__(self, service=None, config=None):
+        """Constructor."""
+        super(RecordVersionsResource, self).__init__(
+            config=self.load_config(config))
+        self.service = service or RecordDraftService()
+
+    def _get_es_preference(self):
```
lnielsen, line 50: Is this function not repeated from somewhere else? perhaps just extract the function and reuse it both places.

invenio_drafts_resources/services/records/service.py
```diff
@@ -280,6 +280,35 @@ def new_version(self, id_, identity, links_config=None):
         return self.result_item(
             self, identity, draft, links_config=links_config)
 
+    def search_versions(self, id_, identity, params=None, links_config=None,
+                        es_preference=None, **kwargs):
+        """Search for record's versions."""
+        record = self.record_cls.pid.resolve(id_, registered_only=False)
+        self.require_permission(identity, "search", record=record)
```
lnielsen, line 287: This has to be ``read`` permission.

lnielsen (approved):
Just two minor things.

invenio_drafts_resources/resources/records/resource.py
zzacharo: I thought to create a `BaseRecordResource` here https://github.com/inveniosoftware/invenio-records-resources/blob/master/invenio_records_resources/resources/records/resource.py#L23 . Should I go for it?

PR #113, opened 2021-03-18
versioning: rename parent_index to index

fenekku:
- closes https://github.com/inveniosoftware/invenio-rdm-records/issues/459

### DIFF
--- a/tests/mock_module/mappings/v6/draftsresources/drafts/draft-v1.0.0.json
+++ b/tests/mock_module/mappings/v6/draftsresources/drafts/draft-v1.0.0.json
@@ -5,8 +5,12 @@
         "id": {
           "type": "keyword"
         },
-        "conceptid": {
-          "type": "keyword"
+        "parent": {
+          "properties": {
+            "id": {
+              "type": "keyword"
+            }
+          }
         },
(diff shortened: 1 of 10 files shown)


tests/mock_module/mappings/v6/draftsresources/drafts/draft-v1.0.0.json
```diff
@@ -39,6 +43,22 @@
         "uuid": {
           "type": "keyword"
         },
+        "versions": {
+          "properties": {
```
lnielsen, line 47: Minor: I think you also need to add ``latest_index`` and ``latest_id``

setup.py
```diff
@@ -59,6 +59,7 @@
 ]
 
 install_requires = [
+    'Flask-SQLAlchemy>=2.1,<2.5.0',  # TEMP: until invenio-db released
```
lnielsen, line 62: Comment: Merged and released as v1.0.9. Thanks a lot.

lnielsen:
In addition to above changes there's also a change to ``schema.py``. In RDM-Records you'll need to change the Alembic recipe as well.

PR #114, opened 2021-03-19
records-versions: redirect to record latest version

zzacharo:
closes inveniosoftware/invenio-app-rdm#707

### DIFF
--- a/invenio_drafts_resources/resources/records/errors.py
+++ b/invenio_drafts_resources/resources/records/errors.py
@@ -1,0 +1,49 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2021 CERN.
+#
+# Invenio-Drafts-Resources is free software; you can redistribute it and/or
+# modify it under the terms of the MIT License; see LICENSE file for more
+# details.
+
+"""Error classes and handlers."""
+
+import json
+
+from flask_resources.errors import HTTPJSONException
+
+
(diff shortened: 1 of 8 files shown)


invenio_drafts_resources/resources/records/errors.py
```diff
@@ -0,0 +1,39 @@
...
+
+    def latest_version_redirected_error_handler(e):
+        response = redirect(e.location, code=301)
+        response.headers['Location'] = e.location
+        return response
+
+    return latest_version_redirected_error_handler
+
+
+class LatestVersionRedirect(Exception):
```
lnielsen, line 25: Minor: Can this not be an [HTTPJSONException](https://github.com/inveniosoftware/flask-resources/blob/master/flask_resources/errors.py#L45)



```python 
def api_redirect(location, code=301):
     raise RedirectException(location, code=code)


class RedirectException(HTTPJSONException):
    code = 301

    def __init__(self, location, **kwargs):
        self.location = location
        kwargs.setdefault(description="Redirecting...")
        super().__init__(**kwargs)
    
    def get_headers(self, environ=None):
        return [("Content-Type", "application/json"), ("Location", self.location)]

    def get_body(self, environ=None):
        """Get the request body."""
        body = {"status": self.code, "message": self.get_description(environ), "location": self.location}
        return json.dumps(body)
```

Then you can completely avoid the error map in the service config.

PR #127, opened 2021-03-24
service: fix issue with reindexing wrong record

lnielsen:


invenio_drafts_resources/services/records/service.py
```diff
@@ -210,8 +210,8 @@ def publish(self, id_, identity, links_config=None):
         self._validate_draft(identity, draft)
 
         # Create the record from the draft
-        record = self.record_cls.publish(draft)
         latest_id = draft.versions.latest_id
+        record = self.record_cls.publish(draft)
```
fenekku, line 214: :1st_place_medal: that must not have been easy to catch and it solved issues I was seeing.

PR #136, opened 2021-04-12
services: validate if files are uploaded on update and publish

audrium:
closes https://github.com/inveniosoftware/invenio-rdm-records/issues/498

### DIFF
--- a/invenio_drafts_resources/services/records/components.py
+++ b/invenio_drafts_resources/services/records/components.py
@@ -8,8 +8,10 @@
 
 """Records service component base classes."""
 
+from flask_babelex import gettext as _
 from invenio_records_resources.services.records.components import \
     MetadataComponent, ServiceComponent
+from marshmallow import ValidationError
 
 
 class PIDComponent(ServiceComponent):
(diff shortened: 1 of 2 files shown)


invenio_drafts_resources/services/records/components.py
```diff
@@ -40,12 +41,23 @@ def new_version(self, identity, draft=None, record=None):
         """New version callback."""
         draft['files'] = record['files']
 
+    def update_draft(self, identity, data=None, draft=None, record=None,
+                     errors=None):
+        """Update callback."""
+        draft_files = record.files
+        if draft_files.enabled and not draft_files.items():
+            errors.append({"field": "files", "messages": [
+                          "Missing uploaded files."]})
```
lnielsen, line 50: 1. The message should be translated:

```
from flask_babelex import gettext as _

_("...")
```

2. The message should be aligned with the one, that's currently in the UI.

invenio_drafts_resources/services/records/service.py
```diff
@@ -161,7 +161,7 @@ def update_draft(self, id_, identity, data, revision_id=None):
         for component in self.components:
             if hasattr(component, 'update_draft'):
                 component.update_draft(
-                    identity, record=draft, data=data)
+                    identity, record=draft, data=data, errors=errors)
```
lnielsen, line 164: Comment: The API change makes sense, however it's likely that there's other components that have to be fixed so just perhaps git grep for ``update_draft()`` in modules to check

audrium: Double checked that, for example `RDMRecordServiceConfig` have in addition to `DraftFilesComponent`:
- MetadataComponent
- AccessComponent
- PIDComponent

all of them have `update_draft` method defined in such a way:
```
def update_draft(self, identity, data=None, record=None, **kwargs)
```
So I guess it's ok

fenekku:
replaced by https://github.com/inveniosoftware/invenio-drafts-resources/pull/138

PR #138, opened 2021-04-13
files: validate files when updating/publishing draft

fenekku:
- requires https://github.com/inveniosoftware/invenio-records-resources/pull/231 to be merged and released
- closes inveniosoftware/invenio-rdm-records#498

### DIFF
--- a/invenio_drafts_resources/services/records/config.py
+++ b/invenio_drafts_resources/services/records/config.py
@@ -17,8 +17,7 @@ from invenio_records_resources.services import \
     SearchOptions as SearchOptionsBase
 from invenio_records_resources.services import pagination_links
 
-from .components import DraftFilesComponent, DraftMetadataComponent, \
-    PIDComponent
+from .components import DraftMetadataComponent, PIDComponent
 from .permissions import RecordPermissionPolicy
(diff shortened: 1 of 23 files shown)


invenio_drafts_resources/services/records/config.py
```diff
@@ -132,6 +133,7 @@ class RecordServiceConfig(RecordServiceConfigBase):
     search_versions = SearchVersionsOptions
 
     components = [
+        DraftFilesComponent,  # Should this be here or just in tests?
```
lnielsen, line 136: Comment: It's related to an issue that Audrius was working on. Currently, RDM-Records is the place where files are married to drafts, and thus 1) the component should actually live there instead or 2) we should move some of the RDM functionality to Drafts-Resources and provide a ``RecordWithFilesService`` or similar.

lnielsen (approved):

invenio_drafts_resources/services/records/config.py
fenekku: Ok, I integrated `DraftFilesComponent` in the test config and am letting rdm-records integrate it.

PR #140, opened 2021-04-20
[WIP] records

### LINKED ISSUE #115: celery: create a task to cleanup deleted drafts
We soft-delete drafts to keep their version id counter around for optimistic concurrency control. They are needed for about 1 min which is the default time for Elasticsearch configured to keep the deleted id around. 

We should create a celery task that can be configured to run once an hour/day, which will look for soft-deleted drafts and hard-delete them. The task should in principle simply call a service method, and should ensure also that e.g. files are removed.

### Implementation details

- Code should be in Invenio-Drafts-Resources in https://github.com/inveniosoftware/invenio-drafts-resources/tree/master/invenio_drafts_resources/services/records
- The task is dealing with the situation documented here: https://github.com/inveniosoftware/invenio-drafts-resources/blob/master/invenio_drafts_resources/records/api.py#L194-L227

A soft-deleted draft happens in two cases (you'll see it from the service methods where ``draft.delete()`` is called:

1. Create a draft, then publish the draft.
2. Create a draft, publish the draft, edit the record, then delete the draft.

#### Test case

https://github.com/inveniosoftware/invenio-drafts-resources/blob/master/tests/services/test_record_service_files.py

```python


draft = service.create(identity_simple, input_data)
record = service.publish(draft.id, identity_simple)

# Here's the task's job (simplified):
draft_model = service.draft_cls.model_cls
draft_model.query.filter(draft_model.is_deleted == True).delete()
db.session.commit()

# Elasticsearch will complain if we try we don't wait a minute, so disable
# the indexer.
monkeypatch.setattr(service.config, 'indexer_cls', MagicMock())

draft = service.edit(record.id, identity_simple)
data = draft.data
data['metadata']['title'] = "New title"
draft = service.update_draft(record.id, identity_simple, data)
record = service.publish(record.id, identity_simple)
```

Overall, this means that we need a method in the service (e.g ``cleanup_drafts()``), which is able to delegate the cleaning of associated objects to the service components (e.g. DraftFilesComponent, should be responsible on edit to recreate the files from the record.

### Tasks

- [x] create the async task to hard delete drafts (with tests) (see [here](https://github.com/inveniosoftware/invenio-indexer/blob/master/invenio_indexer/tasks.py#L19) an example)
- [x] the new async task will have to be added to the Python entrypoints (see [here](https://github.com/inveniosoftware/invenio-indexer/blob/master/setup.py#L94))
- [x] update [invenio-app-rdm](https://github.com/inveniosoftware/invenio-app-rdm/blob/master/invenio_app_rdm/config.py#L246) to add the cron job to CELERY_BEAT

Pineirin:
Develop a method and tests for hard deleting soft deleted drafts after 24 hours since the last update.

Closes #115

### DIFF
--- a/invenio_drafts_resources/records/api.py
+++ b/invenio_drafts_resources/records/api.py
@@ -30,6 +30,7 @@ from invenio_records_resources.records import Record as RecordBase
 from invenio_records_resources.records.systemfields import PIDField, \
     PIDStatusCheckField
 from sqlalchemy.orm.exc import NoResultFound
+import datetime
 
 from .systemfields import ParentField, VersionsField
 
@@ -227,3 +228,10 @@ class Draft(Record):
                 fork_version_id=record.revision_id,
             )
         return draft
+
+    @classmethod
+    def remove_softdeleted(cls, seconds_param):
+        seconds_ago = datetime.datetime.now() - datetime.timedelta(seconds=seconds_param)
+        """the delete() doesn't work do to the foreign key violation problem."""
+        #results = cls.model_cls.query.filter(cls.model_cls.is_deleted == True, cls.model_cls.updated < seconds_ago).delete()
+
--- a/invenio_drafts_resources/services/records/tasks.py
+++ b/invenio_drafts_resources/services/records/tasks.py
@@ -1,0 +1,17 @@
+
+# -*- coding: utf-8 -*-
+#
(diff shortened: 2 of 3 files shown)


Pineirin:
Replaced by https://github.com/inveniosoftware/invenio-drafts-resources/pull/145

PR #148, opened 2021-05-11
Add secret link service 

ines-cruz:
closes  https://github.com/inveniosoftware/invenio-rdm-records/issues/530

### DIFF
--- a/invenio_drafts_resources/services/records/service.py
+++ b/invenio_drafts_resources/services/records/service.py
@@ -25,11 +25,13 @@ class RecordService(RecordServiceBase):
     draft records.
     """
 
-    def __init__(self, config, files_service=None, draft_files_service=None):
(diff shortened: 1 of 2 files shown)


ppanero (commented):
🚀 This simplifies and cancels many of my comments in rdm-records about inheritance... However:

If we define this at this (attribute on the class) at this level, then we should move the `secret_links` service to this module... Othewise we might be inveriting the relationship in dependencies (i.e. invenio-drafts-depends on a class from invenio-rdm-records).

@lnielsen might be able to tell you if:

a) this should be moved to `RDMRecordsService`
b) the `SecretLinksService` should be moved here

My vote (based on the fact that *I think* secret_links is quite rdm dependent... this should be moved up to RDM-Records (a)

ines-cruz:
closed in favor of #https://github.com/inveniosoftware/invenio-rdm-records/pull/595

PR #157, opened 2021-08-03
services: add _get_draft_and_parent_by_id

jrcastro2:
* closes https://github.com/inveniosoftware/invenio-rdm-records/issues/468


lnielsen:
@jrcastro2 Can you have a look at the failing tests.

lnielsen (approved):

lnielsen:
I'm not able to reproduce the test failure on Python 3.9 PostgreSQL 9, ES 6 locally, so I'm simply disabling this matrix case as all other cases are passing, and this one is also passing locally in exactly same environment.

PR #163, opened 2021-09-30
services: Adds post-publish hook

jrcastro2:
* closes https://github.com/inveniosoftware/invenio-rdm-records/issues/563


invenio_drafts_resources/services/records/components/base.py
```diff
@@ -43,3 +43,7 @@ def publish(self, identity, draft=None, record=None):
     def import_files(self, identity, draft=None, record=None):
         """Import files handler."""
         pass
+
+    def post_publish(self, identity, draft=None, record=None):
```
ppanero, line 47: I see all other functions have `draft` in the signature. However, since this happens after it is published... would we need the draft for anything? 🤔 (see we are not even passing it in the call below). What do you think?

jrcastro2: Makes sense :+1:

ppanero (approved):

PR #173, opened 2021-11-09
records: reuse parent record for versions

lnielsen:
* Reuse the parent record when fetching all record versions. Usually we
  get all record versions when we want to reindex all versions (e.g.
  parent was updated). Thus, we inject the parent record into all
  versions, so we avoid having to refetch the parent from the database
  for every record version.


max-moser (approved):
:+1:

lnielsen:
Merged via other PR

Issue #192, opened 2022-04-07
tests: build error on `3.7,pypi,postgresql10,elasticsearch7`


ppanero:
The only packages diff compared to the 3.9 build are (not present or diff version)

3.7 (failing)
```
asttokens==2.0.5
executing==0.8.3
ipython==8.2.0
stack-data==0.2.0
```

3.9 (passing)
```
cached-property==1.5.2
ipython==7.32.0
typing_extensions==4.1.1
```

Failing exception is:

```
[180](https://github.com/inveniosoftware/invenio-drafts-resources/runs/5868837619?check_suite_focus=true#step:7:180)
        bucket = record.bucket
[181](https://github.com/inveniosoftware/invenio-drafts-resources/runs/5868837619?check_suite_focus=true#step:7:181)
>       size_limit = bucket.size_limit
[182](https://github.com/inveniosoftware/invenio-drafts-resources/runs/5868837619?check_suite_focus=true#step:7:182)
E       AttributeError: 'NoneType' object has no attribute 'size_limit'
[183](https://github.com/inveniosoftware/invenio-drafts-resources/runs/5868837619?check_suite_focus=true#step:7:183)
```

Note it does not always fail (e.g. cannot reproduce in local) so it seems to be a timing issue (e.g. something not commited fast enough, etc.)

fenekku:
It fails on 3.9 too, so we can rule out Python version: https://github.com/inveniosoftware/invenio-drafts-resources/actions/runs/5463856950/jobs/9945084100 . Some concurrency/timing issue does seem the more likely problem.

tmorrell:
I think...maybe...this is fixed in the current test setup. We aren't testing on 3.7 anyway, so there can be a new issue is this shows up again.

PR #193, opened 2022-04-29
Include drafts in reindexing of related records

max-moser:
Previously, the reindexing of related records only reindexed sibling *records* (i.e. records sharing the same parent), but not sibling *drafts*.
Thus, the indexed variants of sibling drafts would become outdated when something in their parents changed.

Closes https://github.com/inveniosoftware/invenio-app-rdm/issues/1228


lnielsen (approved):

max-moser:
The test failure looks a lot like https://github.com/inveniosoftware/invenio-drafts-resources/issues/192

PR #196, opened 2022-05-12
service: get related record to reindex when not deleted

### LINKED ISSUE #195: `get_records_by_parent` returns empty dict when no record
`get_records_by_parent` return `[{}]` when the fetched record has no data. It should return `[]` instead.

ntarocco:
* closes #195

### DIFF
--- a/invenio_drafts_resources/records/api.py
+++ b/invenio_drafts_resources/records/api.py
@@ -90,9 +90,13 @@ class Record(RecordBase):
     versions = VersionsField(create=True, set_latest=True)
 
     @classmethod
-    def get_records_by_parent(cls, parent):
+    def get_records_by_parent(cls, parent, include_deleted=True):
         """Get all sibling records for the specified parent record."""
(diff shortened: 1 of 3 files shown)


invenio_drafts_resources/records/api.py
```diff
@@ -90,9 +90,13 @@ class Record(RecordBase):
     versions = VersionsField(create=True, set_latest=True)
 
     @classmethod
-    def get_records_by_parent(cls, parent):
+    def get_records_by_parent(cls, parent, include_deleted=True):
```
ppanero, line 93: I guess is to be backwards compatible, but would we actually use this function with `include_deleted=True`?

ppanero (approved):

max-moser (approved):

invenio_drafts_resources/records/api.py
ntarocco: I don't know that, because the function is at the API level.
The `delete` of the draft, after publish, is happening at service level. So basically, this function doesn't know from where it will be called.

PR #199, opened 2022-06-07
migrate setup and black

utnapischtim:
- migrate setup.py to setup.cfg
- migrate to use black as opinionated auto formater
- add .git-blame-ignore-revs
- fix docs compatibilty problem with Sphinx>=5.0.0
- global: clean test infrastructure


utnapischtim:
the error is not on invenio-drafts-resources side. the error is in invenio-records-resources. how should we handle this? i think to fix this error we would have to create a exception handling in invenio-drafts-resources, and a proper checking of bucket in invenio-records-resources

utnapischtim:
oh and it seams it is a heisenbug! i ran the tests locally and on the first run there where two errors.

invenio_records_resources/records/systemfields/files/manager.py:set_bucket bucket.id NoneType has no Attribute id
invenio_records_resources/services/files/components/content.py:set_file_content bucket.size_limit NoneType has no Attribute size_limit

but after adding some pprint(bucket) statements. the error is gone.

kpsherva (approved):

PR #207, opened 2022-10-12
Fix `preview` permissions: Add flag to skip checking of field-level load permissions in `validate_draft`

max-moser:
UPDATE: I decided to go with the second approach outlined in the other comment, because I feel that avoiding errors is the more sensible approach than catching and ignoring these errors.
The original text for this PR was as follows:

---

The `validate_draft` needs to `schema.load(...)` the (already stored) metadata instead of `schema.dump(...)`, because the latter [does not validate the object to serialize anymore since version 3.0.0rc9](https://github.com/marshmallow-code/marshmallow/blob/dev/src/marshmallow/schema.py#L546-L547).

However, the `RDMRecordSchema` defines [field-level permissions for loading](https://github.com/inveniosoftware/invenio-rdm-records/blob/master/invenio_rdm_records/services/schemas/__init__.py#L45-L47).
This is usually okay because it is typically just the record/draft's creator/owner who can preview the draft.
However, it breaks the share-by-link feature if used to grant `preview` permissions, because the link does not (and should not) give `update_draft` permissions.

So when some person (who is not the owner of the record/draft) accesses the preview page via such a shared link, they will not have the required `update_draft` permission, and the `schema.load(...)` operation will fail due to a `FieldPermissionError` (regardless of whether the draft is valid or not).

This PR fixes that issue by adding a flag that allows to ignore these field-level permissions.
**Note**: This does not circumvent any permission checks, because they are typically already performed elsewhere (e.g. in the service calls used by the view functions' decorators).

Link: [Relevant discussion in Discord](https://discord.com/channels/692989811736182844/704629170470125578/1027950409098219530)

### DIFF
--- a/invenio_drafts_resources/services/records/service.py
+++ b/invenio_drafts_resources/services/records/service.py
@@ -500,12 +500,12 @@ class RecordService(RecordServiceBase):
 
         return ret_val
 
-    def validate_draft(self, identity, id_):
+    def validate_draft(self, identity, id_, ignore_field_permissions=False):
         """Validate a draft."""
         draft = self.draft_cls.pid.resolve(id_, registered_only=False)
-        self._validate_draft(identity, draft)
+        self._validate_draft(identity, draft, ignore_field_permissions)
 
-    def _validate_draft(self, identity, draft):
+    def _validate_draft(self, identity, draft, ignore_field_permissions=False):
         """Validate a draft.
 
         This method is internal because it works with a data access layer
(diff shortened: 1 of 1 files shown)


max-moser:
On second thought... Probably it would be better to prevent the checking of field-level permissions here entirely as opposed to catching the `FieldPermissionError`.
I am afraid that otherwise, the raising of the permission error could potentially mask other errors due to breaking out of normal program control flow.

I'm thinking about setting a no-op `field_permission_check` in the `context` to prevent the construction of the default one in [`ServiceSchemaWrapper._build_context(...)`](https://github.com/inveniosoftware/invenio-records-resources/blob/master/invenio_records_resources/services/records/schema.py#L67)

invenio_drafts_resources/services/records/service.py
```diff
@@ -469,12 +470,12 @@ def rebuild_index(self, identity):
 
         return ret_val
 
-    def validate_draft(self, identity, id_):
+    def validate_draft(self, identity, id_, ignore_field_permissions=False):
```
max-moser, line 473: the flag is set to `False` by default to not break any existing workflows (although i doubt that this method was used anywhere outside of this module)

slint (approved):

PR #208, opened 2022-10-27
Bulk indexing for `rebuild_index`

max-moser:
Make the `service.rebuild_index` method utilize bulk indexing rather than indexing every item individually.
Also add a `draft_indexer` property to the service/config that is separate from the `indexer` property, because each indexer has a set `record_cls` which is used for looking up items to index by ID in bulk indexing.

This PR is related to https://github.com/inveniosoftware/invenio-records-resources/pull/419 ~and currently also includes the changes from https://github.com/inveniosoftware/invenio-drafts-resources/pull/207~.

### DIFF
--- a/invenio_drafts_resources/services/records/config.py
+++ b/invenio_drafts_resources/services/records/config.py
@@ -11,6 +11,7 @@
 """RecordDraft Service API config."""
 
 from flask_babelex import gettext as _
+from invenio_indexer.api import RecordIndexer
 from invenio_records_resources.services import ConditionalLink, RecordLink
 from invenio_records_resources.services import (
     RecordServiceConfig as RecordServiceConfigBase,
@@ -125,6 +126,9 @@ class RecordServiceConfig(RecordServiceConfigBase):
 
     # WHY: We want to force user input choice here.
     draft_cls = None
+
+    draft_indexer_cls = RecordIndexer
+    draft_indexer_queue_name = f"{RecordServiceConfigBase.indexer_queue_name}-drafts"
 
     schema = RecordSchema
 
--- a/invenio_drafts_resources/services/records/service.py
+++ b/invenio_drafts_resources/services/records/service.py
@@ -9,6 +9,7 @@
 
 """Primary service for working with records and drafts."""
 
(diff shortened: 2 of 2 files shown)


invenio_drafts_resources/services/records/config.py
```diff
@@ -126,6 +127,9 @@ class RecordServiceConfig(RecordServiceConfigBase):
     # WHY: We want to force user input choice here.
     draft_cls = None
 
+    draft_indexer_cls = RecordIndexer
```
ppanero, line 130: note to self: mirroring the record indexer cls, see [here](https://github.com/inveniosoftware/invenio-records-resources/blob/master/invenio_records_resources/services/records/config.py#L63)

invenio_drafts_resources/services/records/service.py
```diff
@@ -462,10 +488,8 @@ def rebuild_index(self, identity):
         """
         ret_val = super().rebuild_index(identity)
 
-        for draft_meta in self.draft_cls.model_cls.query.all():
-            draft = self.draft_cls(draft_meta.data, model=draft_meta)
-            if not draft.is_deleted:
-                self.indexer.index(draft)
+        drafts = self.draft_cls.model_cls.query.filter_by(is_deleted=False).all()
+        self.draft_indexer.bulk_index([draft.id for draft in drafts])
```
ppanero, line 492: moderate: same doubt about the records-resources pr, why remove the if?

invenio_drafts_resources/services/records/service.py
```diff
@@ -61,6 +64,29 @@ def draft_cls(self):
         """Factory for creating a record class."""
         return self.config.draft_cls
 
+    @property
+    def draft_indexer(self):
```
ppanero, line 68: minor: would it be worth to abstract this creation and have a `_indexer` factory method that receives the parameter? I think we would not gain much... maybe code dedup?

max-moser: that could very well come in handy, i can have a look at it on monday!

zzacharo (approved):

PR #218, opened 2023-02-12
migrate to flask babel

utnapischtim:
- migrate: flask_babelex replaced by invenio_i18n
- fix: --dry-run ignored

- NOTE: Flask-Security-Invenio invenio-accounts invenio-admin invenio-files-rest invenio-i18n invenio-pidstore invenio-records invenio-records-resources invenio-theme have to be released first

### DIFF
--- a/run-tests.sh
+++ b/run-tests.sh
@@ -52,7 +52,7 @@ if [[ ${keep_services} -eq 0 ]]; then
 fi
 
 python -m check_manifest
-python -m setup extract_messages --dry-run
+python -m setup extract_messages --output-file /dev/null
 python -m sphinx.cmd.build -qnNW docs docs/_build/html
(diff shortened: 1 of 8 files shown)


run-tests.sh
```diff
@@ -52,7 +52,7 @@ if [[ ${keep_services} -eq 0 ]]; then
 fi
 
 python -m check_manifest
-python -m setup extract_messages --dry-run
+python -m setup extract_messages --output-file /dev/null
```
ntarocco, line 55: Can we avoid hard-coding an absolute path? What was the issue with `--dry-run`?

utnapischtim: the Problem with `--dry-run` is that, on the local machine the *.mo or *.po files are regenerated. which could lead to a unwanted commit. the /dev/null should be there on unix like system, but i am not sure on a windows OS. so then we would have to remove that and use the approach from mojib

Issue #219, opened 2023-02-15
files: publishing files pending download from Fetch


TLGINO:
Bug:
When submiting a record through the api and choosing to upload files via the fetch method, it is possible to upload the record before the download (of the file) is done.
This causes lots of bugs down the line.

Related:
https://github.com/inveniosoftware/invenio-drafts-resources/blob/master/invenio_drafts_resources/services/records/components/files.py#L122

https://github.com/inveniosoftware/invenio-records-resources/blob/master/invenio_records_resources/services/files/transfer.py#L44

https://github.com/inveniosoftware/invenio-records-resources/blob/master/invenio_records_resources/services/files/components/metadata.py#L25

PR #220, opened 2023-02-15
files: publishing files pending download from Fetch #219

TLGINO:
* closes https://github.com/inveniosoftware/invenio-drafts-resources/issues/219

### DIFF
--- a/invenio_drafts_resources/services/records/components/files.py
+++ b/invenio_drafts_resources/services/records/components/files.py
@@ -10,6 +10,7 @@
 """Records service component base classes."""
 
 from flask_babelex import gettext as _
+from invenio_records_resources.services.files.transfer import TransferType
 from invenio_records_resources.services.records.components import FilesOptionsComponent
 from marshmallow import ValidationError
 
@@ -130,6 +131,15 @@ class DraftFilesComponent(ServiceComponent):
                     ),
                     field_name="files.enabled",
                 )
+        if draft.files.enabled:
+            for file_record in draft.files.values():
+                if not TransferType(file_record.file.storage_class).is_completed:
+                    raise ValidationError(
+                        _(
+                            "One or more files have not completed their transfer, please wait."
+                        ),
+                        field_name="files",
+                    )
 
         if record.bucket_id:
             self._publish_edit(identity, draft, record)
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -27,6 +27,9 @@ def app_config(app_config):
 
     Needed to set the fields on the custom fields schema.
     """
+    app_config["RECORDS_RESOURCES_FILES_ALLOWED_DOMAINS"] = [
+        "inveniordm.test",
+    ]
     app_config["FILES_REST_STORAGE_CLASS_LIST"] = {
         "L": "Local",
         "F": "Fetch",
--- a/tests/services/test_record_service_files.py
+++ b/tests/services/test_record_service_files.py
@@ -16,11 +16,12 @@ Test to add:
 """
 
 from io import BytesIO
-from unittest.mock import MagicMock
+from unittest.mock import MagicMock, patch
 
 import pytest
 from invenio_files_rest.errors import InvalidOperationError
 from invenio_files_rest.models import Bucket, FileInstance, ObjectVersion
+from invenio_records_resources.services.files.transfer import TransferType
 from marshmallow.exceptions import ValidationError
 from mock_module.models import DraftMetadata, FileDraftMetadata, FileRecordMetadata
 from mock_module.service import ServiceConfig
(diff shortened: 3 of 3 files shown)


invenio_drafts_resources/services/records/components/files.py
```diff
@@ -9,7 +9,10 @@
 
 """Records service component base classes."""
 
+import pydoc
```
ppanero, line 12: Do we need pydoc?

invenio_drafts_resources/services/records/components/files.py
```diff
@@ -64,6 +67,14 @@ def update_draft(self, identity, data=None, record=None, errors=None):
                     ],
                 }
             )
+        elif draft.files.enabled and draft.files.items():
```
ppanero, line 70: If `items()` returns [] then the for loop below will not be entered. There is no need to collect items twice?
```suggestion
        elif draft.files.enabled:
```

invenio_drafts_resources/services/records/components/files.py
```diff
@@ -64,6 +67,14 @@ def update_draft(self, identity, data=None, record=None, errors=None):
...
+            for item in draft.files.items():
+                data = item[1].file.dumps()
```
ppanero, line 72: question: why accesing `[1]`? I'm not a 100% of the types. But I think we could iterate over [`draft.files` (without `.items()`)](https://github.com/inveniosoftware/invenio-rdm-records/blob/ab4390dbcf264b3dc63140e0ad6f1e541747f4b7/invenio_rdm_records/records/api.py#L244). This would return [`FileRecord` objects](https://github.com/inveniosoftware/invenio-records-resources/blob/master/invenio_records_resources/records/api.py#L51), from which we can access the actual [`file` instance](https://github.com/inveniosoftware/invenio-records-resources/blob/master/invenio_records_resources/records/api.py#L71). Then due to the overwrite of [`__getattr__` ](https://github.com/inveniosoftware/invenio-records-resources/blob/master/invenio_records_resources/records/api.py#L171) we can access the uri.

```python
for file_record in draft.files:
   uri = file_record.file.ui
   if not uri.is_completed:
       ...
```

Important: Im describing based on memory and class type assumptions I can be fully wrong.

Reasoning behind all this:
- In general I would try to avoid hardcoding list indixes (i.e. `[1]`) beause they can change and then we won't know we referenced it here and things will break.
- As much as possible avoid using `dumps` and access properties directly. Due to how descriptors work, the properties should be available (if not we might need to check that).

invenio_drafts_resources/services/records/components/files.py
```diff
@@ -64,6 +67,14 @@ def update_draft(self, identity, data=None, record=None, errors=None):
...
+                if not TransferType(data["storage_class"]).is_completed:
+                    raise ValidationError(
+                        _("Upload in progress. " "Please wait."),
```
ppanero, line 75:
```suggestion
                        _("One or more files have not completed its transfer, please wait."),
```

invenio_drafts_resources/services/records/components/files.py
```diff
@@ -64,6 +67,14 @@ def update_draft(self, identity, data=None, record=None, errors=None):
...
+                        field_name="files.enabled",
```
ppanero, line 76: I would put it in the whole files field, I'm not sure if this would display correctly though
```suggestion
                        field_name="files",
```

ppanero (commented):
Nice job! 🚀 I left a few comments but the functionality is fully there.

It might require some mocking, but could we add some tests about this?

invenio_drafts_resources/services/records/service.py
```diff
@@ -527,6 +529,16 @@ def _validate_draft(self, identity, draft):
             # explicit
         )
 
+        if draft.files.enabled:
+            for file_record in draft.files.values():
+                if not TransferType(file_record.file.storage_class).is_completed:
```
ppanero, line 534: To be checked, is `file_record.file.storage_class` already of class TransferType? (check `__class__` since the `__repr__` will show a [string](https://github.com/inveniosoftware/invenio-records-resources/blob/master/invenio_records_resources/services/files/transfer.py#L39))

invenio_drafts_resources/services/records/service.py
```diff
@@ -527,6 +529,16 @@ def _validate_draft(self, identity, draft):
             # explicit
         )
 
+        if draft.files.enabled:
```
ppanero, line 532: major: while in most of the cases records/drafts do contain files, that functionality is optional. For example, vocabularies are "records" that do not contain files. context: What we usually call a _record_ is a _bibliographic record_ (i.e. a publication of some sort).

Since having files is optional, the behaviour is implemented in a component. Meaning, that all the files related behaviour should be in said component. Can we move this check (I think it can be moved as is) here <https://github.com/inveniosoftware/invenio-drafts-resources/blob/master/invenio_drafts_resources/services/records/components/files.py#L133>?

More context: components are configured here https://github.com/inveniosoftware/invenio-drafts-resources/blob/master/invenio_drafts_resources/services/records/config.py#L141 and called here https://github.com/inveniosoftware/invenio-drafts-resources/blob/master/invenio_drafts_resources/services/records/service.py#L341 (you dont have to change any of this)

ppanero (commented):
Very nice job on the test! I think we just need to fix a tiny thing about code placement and we are good to go!

invenio_drafts_resources/services/records/service.py
TLGINO: So doing 
`print(type(file_record.file.storage_class))`
I get
`<class 'str'>
`
So the class instantion is needed no?

invenio_drafts_resources/services/records/service.py
TLGINO: Ok that makes a lot of sense, thanks a lot for the thorough/in depth explanation!
Should be all good now :)

invenio_drafts_resources/services/records/service.py
ppanero: Yup, you are right! thanks for checking!

invenio_drafts_resources/services/records/components/files.py
```diff
@@ -130,6 +131,15 @@ def publish(self, identity, draft=None, record=None):
                     ),
                     field_name="files.enabled",
                 )
+        if draft.files.enabled:
+            for file_record in draft.files.values():
+                if not TransferType(file_record.file.storage_class).is_completed:
```
ppanero, line 136: I think something went wrong int he rebase here. There are two messages (should only have the second one), also is missing the `field_name="files"`

invenio_drafts_resources/services/records/components/files.py
```diff
@@ -130,6 +131,15 @@ def publish(self, identity, draft=None, record=None):
...
+                    raise ValidationError(
+                        (
+                            "One or more files have not completed their transfer, please wait."
+                        ),
+                        field_name="files.enabled",
```
ppanero, line 141:
```suggestion
                        field_name="files",
```

ppanero (approved):

PR #222, opened 2023-02-22
release: v1.0.4

kpsherva:


slint (approved):

ppanero:
Note that the failure on the CI happens arbitrarily (there's an issue for that). It is not reproducible locally, so it is something related to concurrency/update 😞 Should be fixed at some point tho... but it does not seem to be propagated up the chain.

PR #230, opened 2023-06-06
files: pass record to manage files permission check

slint:
Addresses https://github.com/inveniosoftware/invenio-rdm-records/issues/1314

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -9,6 +9,10 @@
 Changes
 =======
 
+Version 1.4.1 (2023-06-06)
+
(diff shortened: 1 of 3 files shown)


slint:
I'm not going to run again the tests on this guy... It seems like they're randomly failing/passing, and looking into it smells like a longer task. Here's proof that the commit including the fix passes tests:

image


I'll push a release commit as well, since it's needed upstream in `invenio-rdm-records` for a similar fix.

alejandromumo (approved):

PR #231, opened 2023-06-22
components: add configurable file components attributes

kpsherva:
* closes https://github.com/inveniosoftware/invenio-app-rdm/issues/2253

# Attention
please read the final comment before reviewing https://github.com/inveniosoftware/invenio-drafts-resources/pull/231#issuecomment-1615840748


invenio_drafts_resources/services/records/components/files.py
```diff
@@ -9,50 +9,50 @@
...
+    MediaFilesOptionsComponent,
+)
+from marshmallow import ValidationError
 
 
-class DraftFilesComponent(ServiceComponent):
+class DraftFilesComponent(RecordFilesComponent):
     """Draft files service component."""
 
+    _files_attr_key = "files"
```
kpsherva, line 24: the actual config of the record file components

invenio_drafts_resources/services/records/components/files.py
```diff
@@ -9,50 +9,50 @@
...
-from .base import ServiceComponent
+from .files_base import RecordFilesComponent
+from invenio_records_resources.services.records.components import (
+    MediaFilesOptionsComponent,
+)
+from marshmallow import ValidationError
 
 
-class DraftFilesComponent(ServiceComponent):
+class DraftFilesComponent(RecordFilesComponent):
```
kpsherva, line 21: all the methods were moved to files_base class (as default implementation/bahaviour for record-file component)

invenio_drafts_resources/services/records/components/files.py
```diff
@@ -9,50 +9,50 @@
...
-                        ],
-                    }
-                )
-                return  # exit early
-
-        draft.files.enabled = enabled
+        files = self.get_record_files(draft)
+
+        # disable the media files by default
+        enabled = data.get(self.files_data_key, {}).get("enabled", False)
```
kpsherva, line 53: in the media files we don't check the manage_files permissions, since media files are additional and they can be added without special restrictions

invenio_drafts_resources/services/records/components/files.py
```diff
@@ -61,33 +61,23 @@ def update_draft(self, identity, data=None, record=None, errors=None):
...
-                        "messages": [
-                            _("You don't have permissions to manage files options.")
-                        ],
-                    }
-                )
-                return  # exit early
+        draft_files = self.get_record_files(draft)
+        record_files = self.get_record_files(record)
+        enabled = data.get(self.files_data_key, {}).get("enabled", record_files.enabled)
+        default_preview = data.get(self.files_data_key, {}).get("default_preview")
```
kpsherva, line 67: same as above

invenio_drafts_resources/services/records/components/files.py
```diff
@@ -103,106 +93,51 @@ def update_draft(self, identity, data=None, record=None, errors=None):
...
+                {
+                    "field": f"f{self.files_data_key}.default_preview",
+                    "messages": e.messages,
+                }
+            )
 
     def edit(self, identity, draft=None, record=None):
         """Edit callback."""
-        if draft.bucket is None:
+        draft_files = self.get_record_files(draft)
```
kpsherva, line 105: this would be better to review side by side with the base files

invenio_drafts_resources/services/records/components/files.py
```diff
@@ -103,106 +93,51 @@ def update_draft(self, identity, data=None, record=None, errors=None):
...
         # TODO: For published records, we should sync changes from the
         # draft bucket to the record bucket, so that an instance could
         # potentially allow a user to update files. For now, sync() only
         # changes the default_preview and order
-        record.files.sync(draft.files)
+        record_files = self.get_record_files(record)
+        draft_files = self.get_record_files(draft)
+        record_files.sync(draft_files)
+        record_files.unlock()
+        record_files.copy(draft_files)
```
kpsherva, line 128: media files are simply copied over from the draft - no new version required - see the RFC

invenio_drafts_resources/services/records/components/files_base.py
```diff
@@ -0,0 +1,239 @@
...
+        """Assigns files.enabled.
+
+        NOTE: `record` actually refers to the draft
+              (this interface is used in records-resources and rdm-records)
+        """
+        draft = record
+        files = self.get_record_files(draft)
+        enabled = data.get(self.files_data_key, {}).get("enabled", True)
+
+        if files.enabled != enabled:
```
kpsherva, line 40: same as before,  we check the permissions

invenio_drafts_resources/services/records/components/files_base.py
```diff
@@ -0,0 +1,239 @@
...
+        draft_bucket = self.get_record_bucket(draft)
+        if draft_bucket is None:
+            # Happens, when a soft-deleted draft is un-deleted.
+            draft[self.files_data_key] = {"enabled": True}
+            # re-fetch the files attribute - above data getter sets the attribute
+            draft_files = self.get_record_files(draft)
+            draft_files.create_bucket()
+        draft_files.copy(record_files)
+        # force going through the new version
+        draft_files.lock()
```
kpsherva, line 127: this part was missing in the regular files - the bucket was not locked, meaning you could edit files in the draft, but they were not copied over to the record - because we require new version for the  regular files changes - resulting in some zombie draft files

invenio_drafts_resources/services/records/components/files_base.py
```diff
@@ -0,0 +1,239 @@
...
+
+    def new_version(self, identity, draft=None, record=None):
+        """New version callback."""
```
kpsherva, line 130: no changes, same as before

invenio_drafts_resources/services/records/components/files_base.py
```diff
@@ -0,0 +1,239 @@
...
+        # We don't copy files from the previous version, but instead allow
+        # users to import the files.
+        draft_files = self.get_record_files(draft)
+        record_files = self.get_record_files(record)
+        draft_files.enabled = record_files.enabled
+
+    def _publish_new(self, identity, draft, record):
+        """Action when publishing a new draft."""
```
kpsherva, line 138: same as before, no changes

invenio_drafts_resources/services/records/components/files_base.py
```diff
@@ -0,0 +1,239 @@
...
+        # changes the default_preview and order
+        record_files = self.get_record_files(record)
+        draft_files = self.get_record_files(draft)
+
+        record_files.sync(draft_files)
+
+        # Teardown the bucket and files created in edit().
+        if draft_files.enabled:
+            draft_files.delete_all()
+        draft_files.unlock()
```
kpsherva, line 173: addition of unlocking the bucket, because of https://github.com/inveniosoftware/invenio-drafts-resources/pull/231/files#r1247649625

tests/services/test_record_service_files.py
```diff
@@ -182,6 +182,39 @@ def test_edit_publish(app, db, service, input_data, identity_simple, monkeypatch
     assert_counts(buckets=1, objs=1, fileinstances=1, filedrafts=0, filerecords=1)
 
 
+def test_edit_draft_files(app, db, service, input_data, identity_simple, monkeypatch):
```
kpsherva, line 185: WIP

invenio_drafts_resources/services/records/components/files_base.py
kpsherva, line 127: @anikachurilova this is where you need to hook in with allowing/disallowing for files updates depending on DOI provenance

invenio_drafts_resources/services/records/components/files_base.py
```diff
@@ -0,0 +1,239 @@
...
+        NOTE: `record` actually refers to the draft
+              (this interface is used in records-resources and rdm-records)
+        """
+        draft = record
+        files = self.get_record_files(draft)
+        enabled = data.get(self.files_data_key, {}).get("enabled", True)
+
+        if files.enabled != enabled:
+            if not self.service.check_permission(
+                identity, "manage_files", record=draft
```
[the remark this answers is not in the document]
kpsherva: media files do not have the "manage_files" permission, if you can edit draft, you can enable/disable media_files too since they are additional, we don't protect them the same way as metadata-only

invenio_drafts_resources/services/records/components/files.py
```diff
@@ -9,50 +9,50 @@
...
+    _files_attr_key = "files"
+    _files_data_key = "files"
+    _files_bucket_attr_key = "bucket"
+    _files_bucket_id_attr_key = "bucket_id"
+
+
+class DraftMediaFilesComponent(RecordFilesComponent):
+    """Draft media files component."""
+
+    _files_attr_key = "media_files"
```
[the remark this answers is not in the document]
kpsherva: do we need to simplify now? isn't it good enough?

kpsherva:
https://github.com/inveniosoftware/invenio-drafts-resources/pull/231#discussion_r1247886483
https://github.com/inveniosoftware/invenio-drafts-resources/pull/231#discussion_r1247891031|

I think I will rename files.py to media_files.py because otherwise there seem to be confusion.
we sync the media files between draft and record since we want to update/modify them without a need of creating a new version (see RFC)

kpsherva:
Screenshot 2023-07-01 at 12 31 30

invenio_drafts_resources/services/records/components/files.py
```diff
@@ -36,23 +41,24 @@ def create(self, identity, data=None, record=None, errors=None):
               (this interface is used in records-resources and rdm-records)
         """
         draft = record
-        enabled = data.get("files", {}).get("enabled", True)
+        files = self.get_record_files(draft)
+        enabled = data.get(self.files_data_key, {}).get("enabled", True)
 
-        if draft.files.enabled != enabled:
+        if files.enabled != enabled:
```
ppanero, line 47: can we join the nested `if`s? The return only happens inside the inner if, so I _think_ the behaviour is left the same.
```suggestion
        if files.enabled != enabled and not self.service.check_permission(
                 identity, "manage_files", record=draft
             ):
```

invenio_drafts_resources/services/records/components/files.py
```diff
@@ -61,16 +67,16 @@ def update_draft(self, identity, data=None, record=None, errors=None):
...
         """
         draft = record
-        enabled = data.get("files", {}).get("enabled", True)
-        default_preview = data.get("files", {}).get("default_preview")
-
-        if draft.files.enabled != enabled:
+        draft_files = self.get_record_files(draft)
+        enabled = data.get(self.files_data_key, {}).get("enabled", True)
+        default_preview = data.get(self.files_data_key, {}).get("default_preview")
+        if draft_files.enabled != enabled:
```
ppanero, line 73: same as above with joining the `if`s

invenio_drafts_resources/services/records/components/files.py
```diff
@@ -81,13 +87,15 @@ def update_draft(self, identity, data=None, record=None, errors=None):
...
+                {"field": f"{self.files_data_key}.enabled", "messages": e.messages}
+            )
             return  # exit early
 
-        if draft.files.enabled and not draft.files.items():
+        if draft_files.enabled and not draft_files.items():
             errors.append(
                 {
-                    "field": "files.enabled",
+                    "field": f"{self.files_data_key}.enabled",
```
ppanero, line 98: Question: How do we know is `data_key` and not `attribute_key`? they are potentially different (according to `FilesOptionsComponent`. However, do we have a way to enforce that generators will use `data_key`?

invenio_drafts_resources/services/records/components/files.py
```diff
@@ -103,76 +111,100 @@ def update_draft(self, identity, data=None, record=None, errors=None):
...
+        draft_files.remove_bucket(force=True)
 
     def publish(self, identity, draft=None, record=None):
         """Copy bucket and files to record."""
-        if draft.files.enabled and draft.files.bucket:
-            if not draft.files.items():
+        draft_files = self.get_record_files(draft)
+        record_bucket_id = self.get_record_bucket_id(record)
+
+        if draft_files.enabled and draft_files.bucket:
```
ppanero, line 189: join the `if`s?

invenio_drafts_resources/services/records/components/media_files.py
```diff
@@ -0,0 +1,138 @@
...
+from invenio_i18n import gettext as _
+from invenio_records_resources.services.records.components import (
+    MediaFilesOptionsComponent,
+)
+from marshmallow import ValidationError
+
+from .files import DraftFilesComponent
+
+
+class DraftMediaFilesComponent(DraftFilesComponent):
```
ppanero, line 21: Maybe is better to make the `MediaFilesOptionsComponent` into mixins that just provide the attribute values. The following suggestion would mean that both inherit `ServiceComponent` (redundant) and otherwise we are redefining attributes that are already defined in records-resources (also redundant)

```suggestion
class DraftMediaFilesComponent(DraftFilesComponent, MediaFilesOptionsComponent):
```

invenio_drafts_resources/services/records/components/media_files.py
```diff
@@ -0,0 +1,138 @@
...
+
+    _files_attr_key = "media_files"
+    _files_data_key = "media_files"
+    _files_bucket_attr_key = "media_bucket"
+    _files_bucket_id_attr_key = "media_bucket_id"
+
+    def __init__(self, service, *args, **kwargs):
+        """Constructor."""
+        super().__init__(service)
+        self.files_component = MediaFilesOptionsComponent(service)
```
ppanero, line 32: I'm not sure I'm understanding the inheritance vs composition here.

In records-resources:

1. [The `BaseFilesOptionsComponent` inherits `ServiceComponent`](https://github.com/inveniosoftware/invenio-records-resources/pull/475/files#diff-c8b3bb7ffe826d4540bf3b1a8c798f2651b5de007ddb43aa367aae13ee5d9e96R20). It defines some attributes and methods and overwrites some methods (update and create)
2. In this module `DraftFilesComponent` inherits `ServiceComponent, BaseRecordFilesComponent`
    - The second (`BaseFilesOptionsComponent`) already inherits `ServiceComponent` so the first addition is redundant
    - `DraftFilesComponent` is redefining all the base `_` attributes to assign them a value. This could be solved by inheriting `FilesOptionsComponent`. However, this would be *wrong* (thus, invalidating my suggestion above), because the  `FilesOptionsComponent` (or children classes) are used as `files_components`.
    - `self.files_component` is only used to call the new defined methods (I do not see any call to the overwritten `create` and `update`.
    
TL;DR: There is a big tangling between composition (i.e. `files_component = FilesOptionsComponent(...)`) and inheritance (i.e. assigning value to `_files_attr_key` and company).

Suggestion: Make `FilesOptionsComponent` (and parent/child class) a mixin. That way, we can inherit and access both attributes and new methods. Without the need to change the consturctor signature.

invenio_drafts_resources/services/records/components/media_files.py
```diff
@@ -0,0 +1,138 @@
...
+        """Assigns files.enabled.
+
+        NOTE: `record` actually refers to the draft
+              (this interface is used in records-resources and rdm-records)
+        """
+        draft = record
+        files = self.get_record_files(draft)
+
+        # disable the media files by default
+        enabled = data.get(self.files_data_key, {}).get("enabled", False)
```
ppanero, line 44: self-check: there is no need to check for permissions as with normal files because if someone has access to the endpoint with media files it is a "power user" of some sort.

invenio_drafts_resources/services/records/components/media_files.py
```diff
@@ -0,0 +1,138 @@
...
+
+        files.enabled = enabled
+
+    def update_draft(self, identity, data=None, record=None, errors=None):
```
ppanero, line 48: This method and the files component `update_draft` are formed 90% the same code, IIUC the only difference is the `enabled` (here is defaults to the record's and in the other component to `True`. Then in the other component, there is a permission check (based on `enabled`), the subsequent code lines 60-92 are the same than lines 95-119. This can be extracted the private methods (one with the enabled and checks over it, which can be different in each component and one for the rest of the logic) and remove duplicated code.

invenio_drafts_resources/services/records/components/media_files.py
```diff
@@ -0,0 +1,138 @@
...
+        draft_bucket = self.get_record_bucket(draft)
+        if draft_bucket is None:
+            # Happens, when a soft-deleted draft is un-deleted.
+            draft[self.files_data_key] = {"enabled": record_files.enabled}
+            # re-fetch the files attribute - above data getter sets the attribute
+            draft_files = self.get_record_files(draft)
+            draft_files.create_bucket()
+        draft_files.copy(record_files)
+        # in the media files we don't lock the bucket
+        # - they can be simply edited from draft
```
ppanero, line 107: This could be a flag on the component and avoid 10 lines of duplicated code.

invenio_drafts_resources/services/records/components/media_files.py
ppanero, line 1: Maybe it can be shelved for a later refactor, but this file has some duplicated code that can be removed using inheritance (see inner comments for more details).

tests/mock_module/api.py
```diff
@@ -64,10 +73,48 @@ class Record(RecordBase):
         delete=False,
     )
 
+    media_files = FilesField(
+        key="media_files",
+        bucket_id_attr="media_bucket_id",
```
ppanero, line 78: I fear this value is duplicated here and then in the `FilesOptionsComponent`, what would happen if there is a miss configuration and the values do not match? I have the gut feeling that they should be somehow centralized (i.e. rely on the systemfield or in the component, but not in both).

tests/mock_module/models.py
```diff
@@ -23,13 +25,26 @@ class DraftMetadata(db.Model, DraftMetadataBase, ParentRecordMixin):
     __tablename__ = "draft_mock_metadata"
     __parent_record_model__ = ParentRecordMetadata
 
+    # bucket_id = db.Column(UUIDType, db.ForeignKey(Bucket.id))
```
ppanero, line 28: remove?

tests/mock_module/models.py
```diff
@@ -23,13 +25,26 @@ class DraftMetadata(db.Model, DraftMetadataBase, ParentRecordMixin):
...
+    # media_bucket = db.relationship(Bucket, foreign_keys=[media_bucket_id])
+
 
 class RecordMetadata(db.Model, RecordMetadataBase, ParentRecordMixin):
     """Model for mock module metadata."""
 
     __tablename__ = "record_mock_metadata"
     __parent_record_model__ = ParentRecordMetadata
 
+    # bucket_id = db.Column(UUIDType, db.ForeignKey(Bucket.id))
```
ppanero, line 40: remove?

tests/services/test_record_service_files.py
ppanero, line 1: shelvable: it is very nice to have tests! 🚀 However, part of the logic relies on records-resources, which does not have said tests. I think we have to split the test suit and move them to where things are defined (without having many duplications). I fear we might change something in records-resources and not see it failing until we run drafts-resources tests.

PR #232, opened 2023-06-30
api: don't override next version with new draft

fenekku:
closes https://github.com/inveniosoftware/invenio-app-rdm/issues/2197

Although small, the fix has a larger context to understand. When a record is published, its draft is soft-deleted in the db until `cleanup_drafts` hard deletes it. When that happens, the published record has no associated draft anymore, so `edit` will create a completely new draft for it. However, that record may have a new version already (this is a fairly common occurrence in a running instance), so the new draft must not overwrite the reference to the `next_draft_id` stored in the parent. Because `next_draft_id` is used both to refer to a record's next version and a never published before draft's own id, we get interference. By checking if we are in the case of a record that has been published (only records that can have a new version) versus a record that has not, we can appropriately set that `next_draft_id`.

Another challenge with this PR was testing it. When trying to test this at the service level, we encounter [OpenSearch's delete behavior](https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-delete.html#delete-versioning) (link to ElasticSearch but it applies to OpenSearch): the deleted document's `version` is kept around for 60s after deletion to overcome concurrency issues: Opensearch keeps a tombstone of deleted documents around and lets you reindex to a deleted document. So the version needs to stick around a little to make sure an update _sent_ just before deletion was processed doesn't create a new record (updates can create new records in ES/OS).

### DIFF
--- a/invenio_drafts_resources/records/systemfields/versions.py
+++ b/invenio_drafts_resources/records/systemfields/versions.py
@@ -209,8 +209,8 @@ class VersionsField(SystemField):
         # The parent record is created on pre_create.
         versions = self.obj(record)
         if self._create and self._set_next:
-            # if fork is none - it's a new?
-            versions.set_next()
+            if not record.is_published:
+                versions.set_next()
         elif self._create and self._set_latest:
             versions.set_latest()
 
--- a/tests/records/test_api.py
+++ b/tests/records/test_api.py
@@ -7,6 +7,7 @@
 # details.
 
 """Data access layer tests."""
+from datetime import timedelta
 
 import pytest
 from invenio_search import current_search_client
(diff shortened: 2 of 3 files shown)


tests/services/test_record_service.py
```diff
@@ -385,3 +385,32 @@ def test_reindexing_all_siblings(app, service, identity_simple, input_data):
...
+
+    Draft.index.refresh()
+
+    # create a new version - can comment/uncomment to try out combinations
+    new_version_draft = service.new_version(identity_simple, recid)
+    assert record._record.versions.next_draft_id == new_version_draft._record.id
+    assert record._record.versions.latest_id == record._record.id
+
+    # create a draft of published record
+    draft = service.edit(identity_simple, record.id)
```
fenekku, line 414: This is the test to look into that exhibits the following behavior:

opensearchpy.exceptions.ConflictError: ConflictError(409, 'version_conflict_engine_exception', '[95283d9d-e29e-4b5f-a9a2-6d6a7b1dac1c]: version conflict, current version [3] is higher than the one provided [1]')
[4:02 PM]

My understanding: The fresh new draft for the existing record is conflicting in the index with the old draft of the same record. This seems to be because cleanup_drafts() doesn't delete the draft document in the index. But that is already strange because when interacting with a local running instance, this conflict error doesn't show up.
[4:02 PM]
So I added self.draft_indexer.delete(d) for each deleted draft d, but that fails with :

NotFoundError(404, '{"_index":"draftsresources-drafts-draft-v1.0.0-1688067005","_id":"95283d9d-e29e-4b5f-a9a2-6d6a7b1dac1c","_version":4,"result":"not_found","_shards":{"total":2,"successful":1,"failed":0},"_seq_no":2,"_primary_term":1}')
 (edited)
[4:02 PM]

So... can't delete because it doesn't exist, but can't index because it does exist :/

invenio_drafts_resources/records/systemfields/versions.py
```diff
@@ -199,8 +199,8 @@ def post_create(self, record):
         # The parent record is created on pre_create.
         versions = self.obj(record)
         if self._create and self._set_next:
-            # if fork is none - it's a new?
-            versions.set_next()
+            if not record.is_published:
```
fenekku, line 202: This is the fix. When interacting with the site in the browser it fixes the issue.

fenekku:
Since the failing tests are due to https://github.com/inveniosoftware/invenio-drafts-resources/issues/192 and may be remedied by https://github.com/inveniosoftware/invenio-drafts-resources/pull/231, I might as well put this up for review.

fenekku:
@Samk13 for review too

tests/records/test_api.py
```diff
@@ -133,6 +133,26 @@ def test_draft_parent_state_hard_delete(app, db, location):
     assert RecordMetadata.query.count() == 0
 
 
+def test_new_draft_of_published_record_doesnt_override_next_draft_id(app, db, location):
+    # Fake a published record without a draft. That's the situation we get after
+    # cleanup_drafts has run on a newly published record
```
zzacharo, line 138: Can you add in the test the "cleanup" action as well so we are sure about the fix?

tests/records/test_api.py
```diff
@@ -133,6 +134,35 @@ def test_draft_parent_state_hard_delete(app, db, location):
     assert RecordMetadata.query.count() == 0
 
 
+def test_new_draft_of_published_record_doesnt_override_next_draft_id(app, db, location):
+
+    # Classic draft of published record having been cleaned up scenario
+    record_1 = Record.publish(Draft.create({}))
+    db.session.commit()
+    Draft.cleanup_drafts(timedelta(0))
```
fenekku, line 142: The most typical case (but relies on cleanup_drafts).

tests/records/test_api.py
```diff
@@ -133,6 +134,35 @@ def test_draft_parent_state_hard_delete(app, db, location):
...
+
+    # Wider test for any situation where draft of record doesn't exist anymore
+    # e.g. admin force deletion
+    parent = ParentRecord.create({})
+    record_2 = Record.create({}, parent=parent)
+    record_2.register()
```
fenekku, line 148: The wider case that covers the above if cleanup_drafts works correctly and covers other scenarios e.g. direct forced deletion etc . Anything that gets us to a published record without a draft.

tests/records/test_api.py
fenekku: done

ntarocco (approved):
Thanks a lot for fixing this nasty one! 👏🏼

PR #237, opened 2023-07-19
Resolve parent IDs

slint:
- Closes https://github.com/inveniosoftware/invenio-rdm-records/issues/1282.
- versions: add parent ID resolver
- api: return generator for record versions

### DIFF
--- a/invenio_drafts_resources/records/systemfields/versions.py
+++ b/invenio_drafts_resources/records/systemfields/versions.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 #
-# Copyright (C) 2021 CERN.
+# Copyright (C) 2021-2023 CERN.
 #
 # Invenio-Drafts-Resources is free software; you can redistribute it and/or
 # modify it under the terms of the MIT License; see LICENSE file for more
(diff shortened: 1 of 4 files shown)


invenio_drafts_resources/records/systemfields/versions.py
```diff
@@ -166,6 +163,19 @@ def __repr__(self):
...
+    """Version field context."""
+
+    @property
+    def model_cls(self):
+        """Versions model class."""
+        return self.record_cls.versions_model_cls
+
+    def resolve(self, *, parent_id=None):
+        """Resolve a versions state from a parent ID."""
+        return self.model_cls.query.filter_by(parent_id=parent_id).one()
```
ntarocco, line 176: should we handle the possible `None` value of the `parent_id` arg? Or will it raise a `NoResultFound` is the value is `None`?

ntarocco (approved):

invenio_drafts_resources/records/systemfields/versions.py
slint: Actually that was a leftover, originally I went a bit ambitious and wanted to also allow passing any version's `record_id` like:

```python
    def resolve(self, *, parent_id=None, record_id=None):
        """Resolve a versions state from a parent ID."""
        assert parnet_id or record_id
        if parent_id:
            return self.model_cls.query.filter_by(parent_id=parent_id).one()
        # ... try to resolve using `record_id`
        ...
```

But there was no real use-case for that in the codebase, so I just left it out. 

I can remove the `=None` default, from the args. Note also that I used the `*` in the definition, so that you always have to pass it as a keyword argument:

```python
# This won't work:
RDMRecord.version.resolve("<some_id>")

# You need to be explicit like so:
RDMRecord.version.resolve(parent_id="<some_id>")
```

PR #244, opened 2023-08-24
tasks: add margin time window for search to purge deleted documents

zzacharo:
Soft deleted records are being kept in Opensearch for [60 seconds by default](https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-delete.html#delete-versioning). That is happening for control of concurrent processes. In our case, when a soft deleted draft is cleaned up but it happened to be soft deleted in less than the time window that Opensearch finally cleans it up, then editing a new record is resulting in version conflict.

This PR adds a new kwarg to deduct the default deletion time and it can be passed as a parameter in case someone has configured the corresponding setting in a different value.

### DIFF
--- a/invenio_drafts_resources/records/api.py
+++ b/invenio_drafts_resources/records/api.py
@@ -22,7 +22,7 @@ identifier.
 """
 
 import uuid
-from datetime import datetime
+from datetime import datetime, timedelta
 
 from invenio_pidstore.models import PIDStatus
 from invenio_pidstore.providers.recordid_v2 import RecordIdProviderV2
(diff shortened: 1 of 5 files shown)


invenio_drafts_resources/records/api.py
```diff
@@ -232,12 +232,19 @@ def edit(cls, record):
...
 
     @classmethod
-    def cleanup_drafts(cls, td):
+    def cleanup_drafts(cls, td, search_gc_deletes=60):
         """Clean up (hard delete) all the soft deleted drafts.
 
         The drafts in the last timedelta span of time won't be deleted.
+        :param int search_gc_deletes: default time search is keeping deleted documents for
+            control of concurrent operations. For more information see:
+            https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-delete.html#delete-versioning
```
ntarocco, line 241:
```suggestion
        :param int search_gc_deletes: time in seconds, corresponding to the search cluster 
          setting `index.gc_deletes` (see
          https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-delete.html#delete-versioning),
          default to 60 seconds. Search cluster caches deleted documents for `index.gc_deletes` seconds.
```

invenio_drafts_resources/records/api.py
```diff
@@ -232,12 +232,19 @@ def edit(cls, record):
         return draft
 
     @classmethod
-    def cleanup_drafts(cls, td):
+    def cleanup_drafts(cls, td, search_gc_deletes=60):
         """Clean up (hard delete) all the soft deleted drafts.
 
         The drafts in the last timedelta span of time won't be deleted.
+        :param int search_gc_deletes: default time search is keeping deleted documents for
```
ntarocco, line 239:
```suggestion
        """Clean up (hard delete) all the soft deleted drafts.

        The soft-deleted drafts in the last timedelta span of time won't be deleted, including
        `search_gc_deletes` seconds timedelta: this ensures that only drafts fully removed 
        from the search cluster can be hard-deleted (e.g. when `td` is very short), avoiding
        search conflicts.
        
        :param int search_gc_deletes: default time search is keeping deleted documents for
```

ntarocco (approved):

PR #246, opened 2023-08-28
Record deletion: Allow passing of 'extra_filter' via kwargs to searches

max-moser:
Previously, the `service.read_drafts()` and `service.read_versions()` methods would not offer a possibility to specify `extra_filter` values that would be taken into account for the created search queries.

This would prevent the RDM-Records service to filter out _deleted records_ (not deleted drafts) from the user's dashboard.

In this PR, I merge the `kwargs["extra_filter"]` (if present) into the default one created in the `search_drafts()` and `read_versions()` methods respectively.
In this way, both the default filter as well as the custom one passed through the keyword arguments apply.

### DIFF
--- a/invenio_drafts_resources/services/records/service.py
+++ b/invenio_drafts_resources/services/records/service.py
@@ -107,6 +107,13 @@ class RecordService(RecordServiceBase):
         # Prepare and execute the search
         params = params or {}
 
+        # `has_draft` systemfield is not defined here. This is not ideal
+        # but it helps avoid overriding the method. See how is used in
+        # https://github.com/inveniosoftware/invenio-rdm-records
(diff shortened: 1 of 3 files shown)


invenio_drafts_resources/services/records/service.py
```diff
@@ -107,19 +107,23 @@ def search_drafts(
         # Prepare and execute the search
         params = params or {}
 
+        # `has_draft` systemfield is not defined here. This is not ideal
```
kpsherva, line 110: why would we try to avoid overridding the method?

max-moser: i have no insights on this comment, i just moved the comment along with the code piece (see L117 on the left)

PR #257, opened 2023-09-29
components: replace ambiguous error on publish

kpsherva:
* when file is being uploaded, if user tries to publish return validation error instead of 500
* closes https://github.com/zenodo/rdm-project/issues/253
* closes https://github.com/inveniosoftware/invenio-app-rdm/issues/2220

# Attention: 
this and dependent PRs need to be reviewed and tested **very** thoroughly

### DIFF
--- a/invenio_drafts_resources/services/records/components/base.py
+++ b/invenio_drafts_resources/services/records/components/base.py
@@ -226,6 +226,19 @@ class BaseRecordFilesComponent(ServiceComponent, _BaseRecordFilesComponent):
         # Teardown the bucket and files created in edit().
         self._purge_bucket_and_ovs(draft_files)
 
+    def _check_file_completed(self, file_record):
+        """Check if file upload has completed."""
+        # Check if RDMDraftFile file has OV assigned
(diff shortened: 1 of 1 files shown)


invenio_drafts_resources/services/records/components/base.py
```diff
@@ -226,6 +226,19 @@ def _publish_edit(self, identity, draft, record):
...
 
+    def _check_file_completed(self, file_record):
+        """Check if file upload has completed."""
+        # Check if RDMDraftFile file has OV assigned
+        # - if not, the upload is ongoing or has failed (fail handled elsewhere)
+        # prevents ambiguous errors when trying to publish a record with
+        # ongoing upload (cannot get storage class of None if OV is not set)
+        has_attached_object = file_record.file is not None
+        if not has_attached_object:
+            return
```
rekt-hard, line 237:
```suggestion
            return False
```
minor: I like to be explicit about return values. None will also evaluate to `False` so feel free to ignore

rekt-hard (approved):

PR #274, opened 2023-11-09
versions: calculate record's active versions

anikachurilova:
* closes https://github.com/zenodo/rdm-project/issues/515

### DIFF
--- a/invenio_drafts_resources/records/api.py
+++ b/invenio_drafts_resources/records/api.py
@@ -27,6 +27,9 @@ from datetime import datetime, timedelta
 from invenio_db import db
 from invenio_pidstore.models import PIDStatus
 from invenio_pidstore.providers.recordid_v2 import RecordIdProviderV2
+from invenio_rdm_records.records.systemfields.deletion_status import (
+    RecordDeletionStatusEnum,
+)
 from invenio_records.systemfields import ModelField
 from invenio_records_resources.records import Record as RecordBase
 from invenio_records_resources.records.systemfields import PIDField, PIDStatusCheckField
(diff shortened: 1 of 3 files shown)


invenio_drafts_resources/records/api.py
```diff
@@ -90,6 +93,14 @@ class Record(RecordBase):
     #: Version relationship
     versions = VersionsField(create=True, set_latest=True)
 
+    @property
+    def versions_count(self):
+        """Get number of record's active versions"""
+        records = list(
+            self.get_record_versions(parent_id=self.parent.id, include_deleted=False)
```
ptamarit, line 100: Could we reuse `get_records_by_parent` with `with_deleted=False, ids_only=True`?

This would fix the tests which are failing due to the import from `invenio_rdm_records` which we cannot do in `invenio-drafts-resources`.
Not sure what is the difference between `is_deleted` and `deletion_status` in the model.

ptamarit (commented):
Should we add a test for `versions_count`?

invenio_drafts_resources/records/api.py
anikachurilova: After checking, I can confirm that we need to use the `deletion_status` and not `is_deleted=False` as `is_deleted` is checking for empty json (see [here](https://github.com/inveniosoftware/invenio-records/blob/master/invenio_records/models.py#L106))

You are right tho, that we can not import `invenio_rdm_records`... I could hardcode the value `"P"` of the `RecordDeletionStatusEnum.PUBLISHED`..

ntarocco (commented):
After checking, we should explore a different implementation.

I would add the property `count` inside the `versions` system field, and calculate the number of versions there.
The query should exclude both records mark as deleted (`RecordDeletionStatusEnum.PUBLISHED.value`) and the soft deleted (`is_deleted=True`).

However, there are a couple of issues:
1. `RecordDeletionStatusEnum.PUBLISHED.value` is in rdm-records and cannot be imported here. Does it make sent to move some of that code here? Or in `records-resources`?
2. the query should be performant, maybe cached, to avoid multiple calculations and performance hits. However, it is not straightforward to understand when invalidating the cache.

PR #279, opened 2024-05-02
versions: dump `latest_id` and `next_draft_id` as string

max-moser:
Dump the UUIDs for `latest_id` and `next_draft_id` as string, if they are not `None`.
Otherwise, the results may not be manageable for further processing (e.g. `json.dumps(...)`).

This is relevant for Invenio-SIPStore, where [`json.dumps(record.dumps())` is used](https://github.com/inveniosoftware/invenio-sipstore/blob/master/invenio_sipstore/api.py#L240)


slint (approved):
LGTM!

Curious about how/where you use `invenio-sipstore` 👀

max-moser:
> Curious about how/where you use invenio-sipstore 👀
In Austria :)

PR #281, opened 2024-05-10
services: fix permission for file edit.

alejandromumo:
closes https://github.com/zenodo/zenodo-rdm/issues/852

### DIFF
--- a/invenio_drafts_resources/services/records/config.py
+++ b/invenio_drafts_resources/services/records/config.py
@@ -36,7 +36,7 @@ def is_record(record, ctx):
 
 
 @staticmethod
-def lock_edit_published_files(service, identity, record=None):
+def lock_edit_published_files(service, identity, record=None, draft=None):
     """Should published files be locked from editing in current record version."""
     return True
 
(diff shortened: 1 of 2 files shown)


invenio_drafts_resources/services/records/config.py
```diff
@@ -36,7 +36,7 @@ def is_record(record, ctx):
 
 
 @staticmethod
```
slint, line 38:
```suggestion
```

nit: I assume `@staticmethod` doesn't do anything in this case

slint (approved):

invenio_drafts_resources/services/records/config.py
utnapischtim: lets have a look at the commit message: https://github.com/inveniosoftware/invenio-drafts-resources/commit/520c951fa47308ad6c1da7291b4c9f9781122c90

slint: I thought that since this goes through the `FromConfig` descriptor (which wasn't the case [at the time of the commit](https://github.com/inveniosoftware/invenio-drafts-resources/commit/520c951fa47308ad6c1da7291b4c9f9781122c90#diff-930c03daf4e4adaee24a69ce17d79e420f3ffed6566aed439badeecf35df067fR157)), `__get__` takes care of the `self` binding part, and just returns a regular function.

But yeah, TIL:

```python
class Proxy:
    def __init__(self, value): 
        self._value = value
    def __get__(self, obj, objtype=None): 
        return self._value

def add(a, b):
    return a + b

class Foo:
    simple_attr_func = add
    descr_attr_func = Proxy(add)

foo = Foo()
foo.simple_attr_func(1, 2)
# raises TypeError: add() takes 2 positional arguments but 3 were given
foo.descr_attr_func(1, 2)
# Returns 3
```

PR #282, opened 2024-05-14
update draft: no restricted records after grace period

anikachurilova:
* closes https://github.com/inveniosoftware/product-rdm/issues/178

### DIFF
--- a/invenio_drafts_resources/services/records/service.py
+++ b/invenio_drafts_resources/services/records/service.py
@@ -258,10 +258,11 @@ class RecordService(RecordServiceBase):
 
     @unit_of_work()
     def update_draft(
-        self, identity, id_, data, revision_id=None, uow=None, expand=False
+        self, identity, id_, data, revision_id=None, uow=None, expand=False, draft=None
     ):
         """Replace a draft."""
-        draft = self.draft_cls.pid.resolve(id_, registered_only=False)
+        if draft:
+            draft = self.draft_cls.pid.resolve(id_, registered_only=False)
 
         self.check_revision_id(draft, revision_id)
 


invenio_drafts_resources/services/records/service.py
```diff
@@ -263,6 +266,21 @@ def update_draft(
...
 
+        # can not make record restricted after grace period
+        if (
+            not current_app.config["ALLOW_RECORD_RESTRICTION_AFTER_GRACE_PERIOD"]
+            and data.get("access", {}).get("record", None) == "restricted"
+            and draft.get("access", {}).get("record", None) == "public"
+        ):
+            end_of_grace_period = (
+                draft.created + current_app.config["RECORDS_RESTRICTION_GRACE_PERIOD"]
+            )
```
kpsherva, line 277: for better conditional statement treadability:
```suggestion
        # can not make record restricted after grace period
        current_draft_is_public = draft.get("access", {}).get("record", None) == "public"
        is_update_to_restricted = data.get("access", {}).get("record", None) == "restricted"
        allow_restriction =  current_app.config["ALLOW_RECORD_RESTRICTION_AFTER_GRACE_PERIOD"]
        
        if not allow_restriction and current_draft_is_public and is_update_to_restricted:
            end_of_grace_period = (
                draft.created + current_app.config["RECORDS_RESTRICTION_GRACE_PERIOD"]
            )
```

invenio_drafts_resources/services/records/service.py
```diff
@@ -263,6 +266,21 @@ def update_draft(
...
         draft = self.draft_cls.pid.resolve(id_, registered_only=False)
 
+        # can not make record restricted after grace period
+        if (
+            not current_app.config["ALLOW_RECORD_RESTRICTION_AFTER_GRACE_PERIOD"]
+            and data.get("access", {}).get("record", None) == "restricted"
+            and draft.get("access", {}).get("record", None) == "public"
+        ):
+            end_of_grace_period = (
+                draft.created + current_app.config["RECORDS_RESTRICTION_GRACE_PERIOD"]
```
kpsherva, line 276: we assume here that grace period is configured, do we want to provide the fallback in case it is not, or to fail? I think the failure is a bit too much user exposed, because it will be only discovered during runtime when draft is being updated. To be discussed.

jrcastro2:
Actually this is not anymore needed as we override the update_draft method in invenio-rdm-records

PR #288, opened 2024-09-16
Services: Records: Raise CommunityNotSelectedError when community is …

sakshamarora1:
closes: https://github.com/CERNDocumentServer/cds-rdm/issues/88

### DIFF
--- a/invenio_drafts_resources/services/records/service.py
+++ b/invenio_drafts_resources/services/records/service.py
@@ -15,6 +15,7 @@ from invenio_pidstore.errors import PIDDoesNotExistError
 from invenio_records_resources.services import LinksTemplate
 from invenio_records_resources.services import RecordService as RecordServiceBase
 from invenio_records_resources.services import ServiceSchemaWrapper
+from invenio_records_resources.services.errors import CommunityNotSelectedError
 from invenio_records_resources.services.uow import (
     RecordBulkIndexOp,
     RecordCommitOp,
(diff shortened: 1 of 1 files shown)


invenio_drafts_resources/services/records/service.py
```diff
@@ -375,6 +376,15 @@ def publish(self, identity, id_, uow=None, expand=False):
...
         # Get the draft
         draft = self.draft_cls.pid.resolve(id_, registered_only=False)
+
+        # If config is True and there are no communities selected
+        # Then, check for permissions to upload without community
+        if (
+            current_app.config["RDM_RECORD_ALWAYS_IN_COMMUNITY"]
+            and len(draft.parent.communities.ids) == 0
+        ):
+            if not self.check_permission(identity, "publish", record=draft):
```
kpsherva, line 386: to improve the readability:
```suggestion

        community_required = current_app.config["RDM_RECORD_ALWAYS_IN_COMMUNITY"]
        community_missing = len(draft.parent.communities.ids) == 0
        can_publish = self.check_permission(identity, "publish", record=draft)
        
        if (
            community_required and community_missing and not can_publish
        ):
```

also not ideal, in some cases we check the permissions for publish twice (second time in line 388)

invenio_drafts_resources/services/records/service.py
```diff
@@ -375,6 +376,15 @@ def publish(self, identity, id_, uow=None, expand=False):
         """
         # Get the draft
         draft = self.draft_cls.pid.resolve(id_, registered_only=False)
+
+        # If config is True and there are no communities selected
+        # Then, check for permissions to upload without community
+        if (
+            current_app.config["RDM_RECORD_ALWAYS_IN_COMMUNITY"]
```
kpsherva, line 383: the configuration variable is declared in rdm-records, therefore this module knows nothing about it if they are not installed together. 
In addition, drafts resources does not know anything about invenio-communities, therefore this check should be done in a place that is aware of both drafts and communities

PR #305, opened 2025-05-14
auditlog_actions: register AuditLogBuilder subclasses to entrypoints

sakshamarora1:
fixes: https://github.com/inveniosoftware/invenio-audit-logs/issues/4


invenio_drafts_resources/auditlog_actions.py
```diff
@@ -8,30 +8,104 @@
...
+        """Update required recipient information and add backend id."""
+        entity_ref = dict_lookup(log_data, self.key)
+        if entity_ref == system_identity.id:
+            entity_data = {
+                "id": str(system_identity.id),
+                "name": "system",
+                "email": "system@system.org",
+            }
+        else:
+            entity = resolver.get_entity_proxy({self.key: entity_ref}).resolve()
```
kpsherva, line 34: this entity proxy should know how to handle system user as well, you shouldn't need the conditional statement in line 27

invenio_drafts_resources/auditlog_actions.py
```diff
@@ -8,30 +8,104 @@
...
+            entity_resolver = context["generator"]
+            entity_resolver(context["resolver"], log_data)
+        return log_data
+
+
+class DraftCreateAuditLog(RecordBaseAuditLog):
+    """Audit log for draft creation."""
+
+    action = "draft.create"
+    message_template = ("User {user_id} created the draft {resource_id}.",)
```
kpsherva, line 83: why is message template a tuple?

invenio_drafts_resources/auditlog_actions.py
```diff
@@ -8,30 +8,104 @@
...
+
+class RecordBaseAuditLog(AuditLogBuilder):
+    """Base class for audit log builders."""
+
+    context = [
+        {
+            "resolver": ServiceResultResolver(service_id="users", type_key="user"),
+            "generator": UserResolve(key="user"),
+        },
+    ]
```
slint, line 51: What about a more unified approach with objects for extracting/building the context:

```python
class RecordAuditLog(AuditLog):

    context = [
        UserContext(),
        RequestContext(),
        TaskContext(),
    ]
    
    ...
```

invenio_drafts_resources/services/records/service.py
```diff
@@ -293,14 +301,7 @@ def update_draft(
...
 
-        current_audit_logs_service.create(
-            data=dict(
-                action="draft.edit",
-                resource=dict(type="record", id=str(id_)),
-            ),
-            identity=identity,
-            uow=uow,
-        )
+        uow.register(AuditLogOp(DraftEditAuditLog.build(id_, identity), identity))
```
slint, line 304:
```suggestion
        uow.register(AuditLogOp(DraftEditAuditLog.build(identity, id_)))
```

We should only keep the `identity` of the "actor" of the audit log event. The identity that creates the audit log is anyways always the system identity.

invenio_drafts_resources/auditlog_actions.py
slint: Figure out how we do translations also.

invenio_drafts_resources/auditlog_actions.py
```diff
@@ -8,30 +8,104 @@
...
+            entity = resolver.get_entity_proxy({self.key: entity_ref}).resolve()
+            entity_data = {
+                "id": str(entity["id"]),
+                "name": entity["username"],
+                "email": entity["email"],
+            }
+        dict_set(log_data, self.key, entity_data)
+
+
+class RecordBaseAuditLog(AuditLogBuilder):
```
utnapischtim, line 43: may i ask why this will be added here and not in `invenio-records-resources`?

invenio_drafts_resources/auditlog_actions.py
```diff
@@ -8,30 +8,119 @@
...
-            message_template="User {user_id} created a new version {resource_id}.",
-        ),
-    }
+from invenio_access.permissions import system_user_id
+from invenio_audit_logs.services import AuditLogAction
+from invenio_records.dictutils import dict_lookup, dict_set
+from invenio_users_resources.entity_resolvers import UserResolver
+
+
+class UserContext:
```
kpsherva, line 17: I think it would be cleaner if these context classes get their own file fe `context.py`
how about giving both a folder `auditlog` and placing both under?

invenio_drafts_resources/auditlog_actions.py
```diff
@@ -8,30 +8,119 @@
...
+            dict_set(data, "metadata.revision_id", record_versions[-1].transaction_id)
+        if self.parent_pid:
+            dict_set(data, "metadata.parent_pid", record.parent.pid.pid_value)
+
+
+class RecordBaseAuditLog(AuditLogAction):
+    """Base class for audit log builders."""
+
+    context = [
+        UserContext(),
```
kpsherva, line 71: @utnapischtim to answer your question in https://github.com/inveniosoftware/invenio-drafts-resources/pull/305#discussion_r2098406467
we have a problem with circular dependency at the moment, because of user context. Normally I would say this class should go to audit logs ... but if we want to use audit logs to log action on users? same dependency issue will come up if we implement this class in records-resources. 
We didn't find a good place yet.

invenio_drafts_resources/auditlog_actions.py
```diff
@@ -8,30 +8,119 @@
...
+from invenio_access.permissions import system_user_id
+from invenio_audit_logs.services import AuditLogAction
+from invenio_records.dictutils import dict_lookup, dict_set
+from invenio_users_resources.entity_resolvers import UserResolver
+
+
+class UserContext:
+    """Payload generator for audit log using the user entity resolver."""
+
+    def __init__(self, key="user"):
```
ntarocco, line 20: question: can the `key` be different? In such case, would the hardcoded UserResolver class work?

invenio_drafts_resources/auditlog_actions.py
```diff
@@ -8,30 +8,119 @@
...
+    def __init__(self, revision_id=False, parent_pid=False):
+        """Ctor."""
+        # Enable or Disable adding metadata
+        self.revision_id = revision_id
+        self.parent_pid = parent_pid
+
+    def __call__(self, data, **kwargs):
+        """Update data with resolved record data."""
+        record = kwargs.get("record", None)
+        if record is None:
```
ntarocco, line 58: should this fail instead?

setup.cfg
```diff
@@ -35,6 +35,7 @@ tests =
     pytest-black-ng>=0.4.0
     pytest-invenio>=3.0.0,<4.0.0
     invenio-app>=2.0.0,<3.0.0
+    invenio-users-resources>=7.0.0,<8.0.0 # Only for tests
```
ntarocco, line 38: why the comment?

invenio_drafts_resources/auditlog_actions.py
ntarocco, line 17: Are the contexts used somehow? I don't see them in the other PRs.

invenio_drafts_resources/auditlog_actions.py
```diff
@@ -8,30 +8,119 @@
...
+        self.key = key
+        self.resolver = UserResolver()
+
+    def __call__(self, data, lookup_key="user_id", **kwargs):
+        """Update data with resolved user data."""
+        entity_ref = dict_lookup(data, lookup_key)
+        entity_proxy = self.resolver.get_entity_proxy({self.key: entity_ref})
+        if entity_ref == system_user_id:
+            entity_data = entity_proxy.system_record()
+            entity_data = {
```
sakshamarora1, line 31: We move the picking of fields needed for audit logs to audit logs, to keep this generic

invenio_drafts_resources/auditlog_actions.py
```diff
@@ -8,30 +8,119 @@
...
+                "id": str(entity_data["id"]),
+                "name": str(entity_data["username"]),
```
sakshamarora1, line 33: change to username inside audit logs user schema

invenio_drafts_resources/auditlog_actions.py
```diff
@@ -8,30 +8,119 @@
...
+                "email": str(entity_data["email"]),
+            }
+        else:
+            entity_data = entity_proxy.resolve()
```
sakshamarora1, line 37: call resolve first and then do type check

invenio_drafts_resources/auditlog_actions.py
utnapischtim: yes, maybe the circular dependency is not fixed, but other functionalities which are using `invenio-records-resources` and not `invenio-drafts-resources` could use this feature too without copy pasting code

invenio_drafts_resources/auditlog/context.py
```diff
@@ -0,0 +1,54 @@
...
+            dict_set(data, "metadata.revision_id", record_versions[-1].transaction_id)
+        if current_app.config["AUDIT_LOGS_METADATA_FIELDS"]["parent_pid"]:
+            dict_set(data, "metadata.parent_pid", record.parent.pid.pid_value)
+
+
+class RequestContext(object):
+    """Payload generator for audit log using the request context."""
+    def __call__(self, data, **kwargs):
+        """Update data with resolved request data."""
+        # IMPORTANT! DON'T COPY THIS, PLEASE DON'T DO THIS EVER...
```
kpsherva, line 49: 😅 can you add a line below explaining why we shouldn't copy?

zzacharo (approved):

PR #316, opened 2025-11-04
tests: set transition range for python versions (rdm)

fenekku:


tmorrell:
I think you want to disable media files here: https://github.com/inveniosoftware/invenio-drafts-resources/blob/95e7ca19705c9039fcfc62712446e1494dfc559b/tests/conftest.py#L80

PR #317, opened 2025-11-07
refactor: remove deprecated usage of Link (and derivatives)

fenekku:


invenio_drafts_resources/services/records/config.py
```diff
@@ -164,26 +167,33 @@ class RecordServiceConfig(RecordServiceConfigBase):
...
-            if_=RecordLink("{+api}/records/{id}"),
-            else_=RecordLink("{+api}/records/{id}/draft"),
+            if_=RecordEndpointLink("records.read"),
+            else_=RecordEndpointLink("records.read_draft"),
         ),
         "self_html": ConditionalLink(
             cond=is_record,
-            if_=RecordLink("{+ui}/records/{id}"),
-            else_=RecordLink("{+ui}/uploads/{id}"),
+            if_=RecordEndpointLink("invenio_app_rdm_records.record_detail"),
```
utnapischtim: as i understand the `RecordEndpointLink` `invenio_app_rdm_records` is [this](https://github.com/inveniosoftware/invenio-app-rdm/blob/master/invenio_app_rdm/records_ui/views/__init__.py#L72)  blueprint. which means that we add an implicit dependency to `invenio-app-rdm` here.

in my opinion this is not good. i thought `invenio-drafts-resources` is independent of `app-rdm`.

invenio_drafts_resources/services/records/config.py
```diff
@@ -164,26 +167,33 @@ class RecordServiceConfig(RecordServiceConfigBase):
     links_item = {
         "self": ConditionalLink(
             cond=is_record,
-            if_=RecordLink("{+api}/records/{id}"),
-            else_=RecordLink("{+api}/records/{id}/draft"),
+            if_=RecordEndpointLink("records.read"),
```
utnapischtim: where is `records.read` defined? i understand `invenio_app_rdm_records`below is a blueprint, but i can't find a `records` blueprint

utnapischtim: is it [this](https://github.com/inveniosoftware/invenio-rdm-records/blob/master/invenio_rdm_records/resources/config.py#L239)?

but that would mean, we have the same problem as below with `invenio_app_rdm_records`, we introduce a implicit dependency to `invenio-rdm-records

fenekku:
@utnapischtim the 2 points above are about the same thing so I will explain it here.

To cut to the chase, the *existing* (prior) behavior *was* coupled to invenio-app-rdm and rdm-records and, IMHO, in a much more insidious way. The change makes it explicit what it expects to be implemented and solves some of the problems. The changes stick to what was there already to keep the same interface and not introduce a change in interface (however obscure) so no major bump needed for a task that was really about cleaning up deprecation warnings.

What already existed is the generation of `/records/<pid_value>`... links in the API response whose views were actually implemented by invenio-app-rdm/invenio-rdm-records only (without those modules, anything using invenio-drafts-resources would be generating links to nowhere). Also, invenio-app-rdm [has configs](https://github.com/inveniosoftware/invenio-app-rdm/blob/master/invenio_app_rdm/config.py#L797) that was letting the routes to those views change without link generation in invenio-drafts-resources ever knowing. Using the `RecordEndpointLink` and co. at least makes those changes carry back.

To be fair, in practice, invenio-rdm-records completely [overrides the config](https://github.com/inveniosoftware/invenio-rdm-records/blob/master/invenio_rdm_records/services/config.py#L645) which really means that what was there or the change I made (which keeps the existing interface) are irrelevant anyway for classic InvenioRDM usage. 

The outer default interface is kept just to keep same interface and not go through all that major bump process... I do think invenio-drafts-resources **should get out of the business of providing default links** though :smile_cat:. It makes no sense since only concrete services, attached to concrete registered resources do implement endpoints anyway. invenio-drafts-resources in my mind is more a series of abstract classes that concrete ones can inherit from. Like invenio-records-resources does: https://github.com/inveniosoftware/invenio-records-resources/blob/master/invenio_records_resources/services/records/config.py#L78

utnapischtim (approved):

PR #323, opened 2026-02-25
refactor(auditlog): Define metadata schema dynamically

sakshamarora1:
requires: https://github.com/inveniosoftware/invenio-audit-logs/pull/21
closes: https://github.com/inveniosoftware/invenio-audit-logs/issues/23

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -11,6 +11,12 @@
 Changes
 =======
 
+Version v9.0.0 (released 2026-04-02)
+
+- refactor(auditlog): RequestContext using X-Forwarded-for & flask session
+- refactor(auditlog): Add parent_pid to draft actions
+- refactor(auditlog): Define metadata schema dynamically
+
 Version v8.0.1 (released 2026-03-10)
 
 - fix(api): Cleanup next_draft_id in versions state only on latest draft
--- a/invenio_drafts_resources/__init__.py
+++ b/invenio_drafts_resources/__init__.py
@@ -11,6 +11,6 @@
 
 """Invenio Drafts Resources module to create REST APIs."""
 
-__version__ = "8.0.1"
+__version__ = "9.0.0"
 
 __all__ = ("__version__",)


invenio_drafts_resources/auditlog/context.py
```diff
@@ -49,14 +55,10 @@ def __call__(self, data, **kwargs):
...
-                )
-                if session:
-                    dict_set(data, "metadata.session", session)
+            ip = request.headers.get("REMOTE_ADDR") or request.remote_addr
+            if ip:
+                dict_set(data, "metadata.ip_address", ip)
+
+            session = request.cookies.get("SESSION") or request.cookies.get("session")
+            if session:
+                dict_set(data, "metadata.session", session)
```
OliverGeneser: ip_address and session is required fields in the metadata schema

invenio_drafts_resources/auditlog/context.py
```diff
@@ -35,11 +35,17 @@ class RecordContext(object):
...
+        record_versions = record.model.versions.all()
+        dict_set(data, "metadata.revision_id", record_versions[-1].transaction_id)
+
+
+class ParentContext(object):
+    """Payload generator for audit log to get parent data."""
+
+    def __call__(self, data, **kwargs):
+        """Update data with resolved parent data."""
+        parent = kwargs["parent"]
```
OliverGeneser: nit; KeyError if we forget to add parent in one of the call sites

invenio_drafts_resources/auditlog/context.py
```diff
@@ -35,11 +35,17 @@ class RecordContext(object):
     def __call__(self, data, **kwargs):
         """Update data with resolved record data."""
         record = kwargs["record"]
-        if current_app.config["AUDIT_LOGS_METADATA_FIELDS"]["revision_id"]:
-            record_versions = record.model.versions.all()
-            dict_set(data, "metadata.revision_id", record_versions[-1].transaction_id)
-        if current_app.config["AUDIT_LOGS_METADATA_FIELDS"]["parent_pid"]:
-            dict_set(data, "metadata.parent_pid", record.parent.pid.pid_value)
+        record_versions = record.model.versions.all()
```
OliverGeneser: I don't really like N+1... Is it possible to order_by(date, id, etc.) and then limit(1)? So we only return the latest

invenio_drafts_resources/auditlog/context.py
```diff
@@ -49,14 +55,10 @@ def __call__(self, data, **kwargs):
         """Update data with resolved request data."""
         # IMPORTANT! DON'T COPY THIS, PLEASE DON'T DO THIS EVER...
         if request:
```
OliverGeneser: I know you didn't change it, but isn't request always truthy in Flask??? Shouldn't we check if the context is available? https://flask.palletsprojects.com/en/stable/api/#flask.has_request_context

invenio_drafts_resources/auditlog/context.py
sakshamarora1: It's not possible to do filtering and ordering on the sqlalchemy-continuum versioned model here. I did try the following but the descending order doesn't work probably because of the default ordering
```python
record.model.versions.order_by(RDMRecordMetadataVersion.transaction_id.desc()).first()
```

`record.model.versions._order_by_clauses` gives:
```
(Column('transaction_id', BigInteger(), table=<rdm_records_metadata_version>, primary_key=True, nullable=False),)
```

I agree it's not the best. If it makes things clearer, I can add a comment that the versioned class is by default ordered by the transaction id.

OliverGeneser: Does this work?

```
# will erase all ORDER BY and ORDER BY new_col alone
q = q.order_by(None).order_by(new_col)
```
https://docs.sqlalchemy.org/en/14/orm/query.html

sakshamarora1: Yes, thanks! Why didn't I try that 😆
```python
record.model.versions.order_by(None).order_by(desc('transaction_id')).first()
```

invenio_drafts_resources/auditlog/context.py
```diff
@@ -48,15 +63,9 @@ class RequestContext(object):
...
-                    "session"
-                )
-                if session:
-                    dict_set(data, "metadata.session", session)
+        if has_request_context():
+            ip = request.headers.get("REMOTE_ADDR") or request.remote_addr
+            dict_set(data, "metadata.ip_address", ip)
+
+            session = request.cookies.get("SESSION") or request.cookies.get("session")
+            dict_set(data, "metadata.session", session)
```
zzacharo: This doesnt seem the best approach for safety:
- Headers and cookies can be manipulated by the user (we need to use safer options like flask.session and  maybe our trusted proxy e.g `X-FORWARDED-FOR` header), ping @slint @ntarocco @kpsherva 
- storing the IP: have we evaluated the potential impact of this in the Privacy notice of a RDM instance?

invenio_drafts_resources/auditlog/context.py
```diff
@@ -48,15 +63,9 @@ class RequestContext(object):
     def __call__(self, data, **kwargs):
         """Update data with resolved request data."""
         # IMPORTANT! DON'T COPY THIS, PLEASE DON'T DO THIS EVER...
```
ntarocco: What is this comment?

invenio_drafts_resources/auditlog/context.py
```diff
@@ -48,15 +63,9 @@ class RequestContext(object):
...
-                    dict_set(data, "metadata.ip_address", ip)
-
-            if current_app.config["AUDIT_LOGS_METADATA_FIELDS"]["session"]:
-                session = request.cookies.get("SESSION") or request.cookies.get(
-                    "session"
-                )
-                if session:
-                    dict_set(data, "metadata.session", session)
+        if has_request_context():
+            ip = request.headers.get("X-Forwarded-For", request.remote_addr)
```
ntarocco: in flask-security, it has been implemented in a [slightly different way](https://github.com/pallets-eco/flask-security-3.0/blob/4049c0620383f42d37950c7a35af5ddd6df0540f/flask_security/utils.py#L65), can you check?

invenio_drafts_resources/auditlog/context.py
ntarocco: For the X-Forwarded-For, it should be controlled by the web server + the Flask ProxyFix. It should be safe to use.
Regarding storing the IP, it should be done only for audit logs that requires that. Is that the case @sakshamarora1?
If yes, then there is a legitimate purpose, and it should be added to the privacy policy (this should be part of the InvenioRDM documentation). The retention period should match the institution's policy.

sakshamarora1: The IP is stored for all draft actions as of right now but it is a context resolver and can be removed for an action when setting up the entrypoint

invenio_drafts_resources/auditlog/context.py
sakshamarora1: Thank you!

invenio_drafts_resources/auditlog/context.py
sakshamarora1: Also, I checked the [Documentation](https://werkzeug.palletsprojects.com/en/stable/middleware/proxy_fix/) for ProxyFix, and it says there that `X-Forwarded-For sets REMOTE_ADDR`
(With the flask-security handling linked, the if statement is there to take care of non-ProxyFix implementations)
https://stackoverflow.com/a/51088317

sakshamarora1: Updated the PR, with using the `X-forwarded-For`! Thank you for pointing this out.

